PR #3, opened 2020-02-27
#2 fix the autoloading of mollie api

### LINKED ISSUE #2: Integration via Composer not possible
Hello everybody, 

can someone briefly explain to me why the autoload of the API is also listed here? Actually every composer dependency should handle its own autoload.

https://github.com/mollie/Shopware6/blob/a6a88f0eb7a61a389f099f8e5b348221c380cc56/composer.json#L33

As a result, the plugin cannot be integrated via Composer, because the plugin path no longer fits here.

henrikvolmer:
Fixes #2

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,7 +1,7 @@
 {
   "name": "kiener/mollie-payments-plugin",
   "description": "Mollie Payments",
-  "version": "v1.0.3",
+  "version": "v1.0.4",
   "type": "shopware-platform-plugin",
   "license": "MIT",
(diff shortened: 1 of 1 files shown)


KienerNL:
I did some fixes for the autoloading. Can removing the autoload psr-4 from the composer.json doesn't seem to work. I did add optimize-autoload that might help. And I also added a require once for the vendor/autoload.php. Can you please check if this works out for you? Thanks for helping out!

PR #7, opened 2020-04-03
Mollie is performing POST requests to the webhook url

coudenysj:


coudenysj:
Looks like this fix is not complete. I now get this error:

```
Uncaught PHP Exception Shopware\Storefront\Framework\Csrf\Exception\InvalidCsrfTokenException: "The provided CSRF token is not valid"
```

KienerNL:
The webhook controller can now receive POST requests without the csrf being checked. Thanks for your input.

coudenysj:
What was wrong with this MR? :thinking:

Issue #23, opened 2020-05-27
LoggerService doesn't limit message length


soebbing:
Hi there!

The `Loggerservice` doesn't enforce the maximum message length defined in the `log_entry` table. 

Messages can only be 255 chars long, this length exceeded during checkout with the klarna "Slice It" payment in our case (the payment wasn't available in our country, resulting in a longer error message).

Changing the line 54 in the `LoggerService` by adding a `mb_substr()` call, reducing the length to 255 chars fixed the issue in our case.

reindernl:
Hi Soebbing,

Thanks for contacting us. We actually found this issue ourselves already and fixed it with a mb_substr 😊

My kind regards,
Reinder.

PR #24, opened 2020-05-27
Billing country always "NL"

soebbing:
When testing the Klarna Slice It. payment, I discovered that the Mollie API responded with: "The billing country is not supported for the payment method".

I was using a German address, that payment should be possible there. During debugging I realized the plugin send "NL" as the billing country, coming from the default fallback when no country was provided.

The country wasn't available, because it's association wasn't loaded initially. The Shopware DAL only loads OneToOne associations by default, and this is also true for associations of associations (like the country in an address).

So this fix adds the relations for the country to all loaded addresses.
image

And I also added a commit for the issue #23 that occured during debugging as a bonus :)

### DIFF
--- a/src/Service/CustomerService.php
+++ b/src/Service/CustomerService.php
@@ -71,10 +71,12 @@ class CustomerService
         try {
             $criteria = new Criteria();
             $criteria->addFilter(new EqualsFilter('id', $customerId));
-            $criteria->addAssociation('activeShippingAddress');
-            $criteria->addAssociation('activeBillingAddress');
-            $criteria->addAssociation('defaultShippingAddress');
-            $criteria->addAssociation('defaultBillingAddress');
+            $criteria->addAssociations([
+                'activeShippingAddress.country',
+                'activeBillingAddress.country',
+                'defaultShippingAddress.country',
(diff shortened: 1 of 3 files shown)


src/Service/CustomerService.php
```diff
@@ -71,10 +71,16 @@ public function getCustomer(string $customerId, Context $context) : ?CustomerEnt
         try {
             $criteria = new Criteria();
             $criteria->addFilter(new EqualsFilter('id', $customerId));
-            $criteria->addAssociation('activeShippingAddress');
-            $criteria->addAssociation('activeBillingAddress');
-            $criteria->addAssociation('defaultShippingAddress');
-            $criteria->addAssociation('defaultBillingAddress');
+            $criteria->addAssociations([
+                'activeShippingAddress',
```
shyim, line 75: You dont need activeShippingAddress when you load activeShippingAddress.country

soebbing: Thanks for pointing it out! Has been fixed! :blue_heart:

reindernl:
Thanks for noticing. I'll accept the pull request!

PR #64, opened 2021-01-13
fixing UTC default

brunogoossens:
DateTimeZone::UTC is not valid and can throw an error: DateTimeZone::__construct(): Unknown or bad timezone (1024)

more info:
https://stackoverflow.com/questions/34839218/php-strange-contants-in-datetimezone-class


milan-kiener:
Hi Bruno, 

Unfortunately, the master branch is a severely outdated version and should not be used. The issue you are trying to fix has already been resolved since November 19th / version 1.3.7

PR #69, opened 2021-02-08
FIX: ContextPersister requires saleschannelId in 6.3.4

milan-kiener:
This fixes an issue with Apple Pay Direct and the upcoming PayPal Shortcut in 6.3.5+

### DIFF
--- a/src/Service/CustomerService.php
+++ b/src/Service/CustomerService.php
@@ -44,15 +44,19 @@ class CustomerService
     /** @var EntityRepositoryInterface */
     private $salutationRepository;
 
+    /** @var string */
+    private $shopwareVersion;
+
     /**
      * Creates a new instance of the customer service.
      *
-     * @param EntityRepositoryInterface    $countryRepository
-     * @param EntityRepositoryInterface    $customerRepository
-     * @param EventDispatcherInterface     $eventDispatcher
-     * @param LoggerInterface              $logger
+     * @param EntityRepositoryInterface $countryRepository
+     * @param EntityRepositoryInterface $customerRepository
+     * @param EventDispatcherInterface $eventDispatcher
+     * @param LoggerInterface $logger
      * @param SalesChannelContextPersister $salesChannelContextPersister
-     * @param EntityRepositoryInterface    $salutationRepository
+     * @param EntityRepositoryInterface $salutationRepository
+     * @param string $shopwareVersion
      */
     public function __construct(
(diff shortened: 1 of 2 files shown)


barbieswimcrew (approved):

src/Service/CustomerService.php
```diff
@@ -100,7 +100,9 @@ public function customerLogin(CustomerEntity $customer, SalesChannelContext $con
                 'customerId' => $customer->getId(),
                 'billingAddressId' => null,
                 'shippingAddressId' => null,
-            ]
+            ],
```
schmitzcarsten, line 103: I think this will break old shopware versions. The method signature of the save function does not have the new 2 parameters (checked shopware 6.1)
Would you please check this? And if I'm right, please implement a solution that works in all shopware versions

schmitzcarsten (changes requested):

src/Service/CustomerService.php
milan-kiener: It should be compatible with older versions of shopware* because it's always allowed to pass in extra parameters even if a function does not expect them, you could still be requesting the extra params using `func_get_args()` after all. Yes your IDE will whine about it, but it doesn't actual break anything.

The only problematic version, hence the asterisk, is 6.3.3.x. The order of parameters for the context persister is:
- In 6.3.2 and lower: token, parameters.
- In 6.3.3: token, parameters, **customerId**
- In 6.3.4 and higher: token, parameters, salesChannelId, **customerId**

The only thing I can think of to include compatibility with 6.3.3, is to have an if else where we check for the same feature flag as shopware does in the context persister, which is NEXT_10058. But it's not recommended to ship plugins that rely on feature flags...

schmitzcarsten: @milan-kiener ok, good that we found this issue. The problem here is, that if we would leave the code as it is, it will break shopware 6.3.3.x. 
So we need a solution to recognize which shopware version is used and if 6.3.3.x we have to call the method with other parameter order.
I agree that we shouldn't use feature flags. So we have to find out which shopware version is used.

schmitzcarsten: @milan-kiener perhaps adding shopware version per DI solves the problem. 
<argument>%kernel.shopware_version%</argument> should do the trick. But I don't know if we get all minor versions with this.
If so, we may use the proposed if switch here and are safe in all shopware versions

schmitzcarsten (approved):

PR #71, opened 2021-02-10
MOL-142: tax calculation for mollie api

schmitzcarsten:
Rounding and tax calculation issue with shopware approach. There are differences in tax calculation in Shopware an Mollie API. There also two implemented ways (horizontal and vertical tax calculation) in Shopware. Because of the different approaches to tax calculation, that may result in different tax amounts in edge cases. 
This PR ensures tax recalculation with the api algorithm for each lineitem that is posted to mollie api.


barbieswimcrew (approved):

src/Service/ApiTaxCalculator.php
```diff
@@ -0,0 +1,13 @@
...
+
+
+namespace Kiener\MolliePayments\Service;
+
+
+class ApiTaxCalculator
+{
+    public function calculateTaxAmount(float $totalAmount, float $vatRate): float
+    {
+        return round($totalAmount, 2) * ($vatRate / (100 + $vatRate));
```
boxblinkracer, line 11: aehm...wouldnt the vatAmount be 3 digits? maybe round that end result one more time?
`24.99 * (16 / 116) = 3,4468....`

src/Service/OrderService.php
```diff
@@ -124,36 +142,29 @@ public function getOrderLinesArray(OrderEntity $order): array
...
+            } catch (MissingPriceLineItemException $e) {
+                $this->logger->critical(
+                    sprintf(
+                        'The order could not be prepared for mollie api. The LineItem with id (%s), sku (%s) has no prices',
+                        $item->getId(),
+                        (string)$sku
+                    )
+                );
+
+                throw new MissingPriceLineItemException($item->getId());
```
boxblinkracer, line 165: great exception handling, bedankt

boxblinkracer: but...you could just throw $e....would be more correct?

src/Service/OrderService.php
```diff
@@ -179,31 +190,15 @@ public function getOrderLinesArray(OrderEntity $order): array
...
                 'name' => $item->getLabel(),
                 'quantity' => $item->getQuantity(),
-                'unitPrice' => $this->getPriceArray($currencyCode, $unitPrice),
-                'totalAmount' => $this->getPriceArray($currencyCode, $totalAmount),
-                'vatRate' => number_format($vatRate, 2, '.', ''),
-                'vatAmount' => $this->getPriceArray($currencyCode, $vatAmount),
+                'unitPrice' => $molliePreparedApiPrices['unitPrice'],
+                'totalAmount' => $molliePreparedApiPrices['totalAmount'],
+                'vatRate' => $molliePreparedApiPrices['vatRate'],
+                'vatAmount' => $molliePreparedApiPrices['vatAmount'],
```
boxblinkracer, line 201: i think it would be more reliable to have a struct object instead of array...

src/Service/OrderService.php
```diff
@@ -288,6 +283,53 @@ public function getShippingItemArray(OrderEntity $order): array
...
+            $vatRate = 0.0;
+        }
+
+        $vatAmount = $this->apiTaxCalculator->calculateTaxAmount($lineItemTotalPrice, $vatRate);
+
+        return [
+            'unitPrice' => $this->getPriceArray($currencyCode, $unitPrice),
+            'totalAmount' => $this->getPriceArray($currencyCode, $lineItemTotalPrice),
+            'vatAmount' => $this->getPriceArray($currencyCode, $vatAmount),
+            'vatRate' => number_format($vatRate, 2, '.', '')
```
boxblinkracer, line 329: attention, number_forma does an implicit rounding or truncation (i forgot) but it onece screwed up something :D

src/Service/OrderService.php
```diff
@@ -288,6 +283,53 @@ public function getShippingItemArray(OrderEntity $order): array
...
+            $unitPrice *= ((100 + $vatRate) / 100);
+            $lineItemTotalPrice += $taxCollection->getAmount();
+        }
+
+        // Remove VAT if the order is tax free
+        if ($orderTaxType === CartPrice::TAX_STATE_FREE) {
+            $vatRate = 0.0;
+        }
+
+        $vatAmount = $this->apiTaxCalculator->calculateTaxAmount($lineItemTotalPrice, $vatRate);
```
boxblinkracer, line 323: i know the function is internally rounding the line item total price, problem is
that in line 327 it would be used without any rounding (or with number_format which might only truncate)

so i would rather see an explicit rounding here in 322 of all base values
then maybe the vatAmount rounding, and these values should be inserted below 

that code looks like sw5 which had these bugs and had to be changed

also if its just 1 line, i would think about removing the apiTaxCalculator function and only adding it here within a "transparent" algorithm and workflow

tests/Fake/LoggerFake.php
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Kiener\MolliePayments\Tests\Fake;
+
+
+use Psr\Log\LoggerInterface;
+
+class LoggerFake implements LoggerInterface
+{
+
+    public function emergency($message, array $context = array())
```
boxblinkracer, line 11: i think there would be a ready to use NullLogger....just to let you know

tests/Service/OrderServiceTest.php
```diff
@@ -0,0 +1,123 @@
...
+     * @param string $orderTaxType
+     * @param string $currencyCode
+     * @param OrderLineItemEntity $lineItem
+     * @dataProvider getVatAmountTestData
+     */
+    public function testRecalculationTaxAmount(string $expected, string $orderTaxType, string $currencyCode, OrderLineItemEntity $lineItem): void
+    {
+        $mollieApiValues = $this->orderService->hydrateLineItemPriceData($lineItem, $orderTaxType, $currencyCode);
+        $actual = $mollieApiValues['vatAmount']['value'];
+        $this->assertSame($expected, $actual);
```
boxblinkracer, line 58: really really down to the core...but i think value is based on string, where "same" might not be what you want to check? maybe more "equals"?

tests/Service/OrderServiceTest.php
```diff
@@ -0,0 +1,123 @@
...
+                CartPrice::TAX_STATE_FREE,
+                'EUR',
+                $this->createOrderLineItemEntity(1, 129.79, 129.79, 0.0, 7.0, CartPrice::TAX_STATE_FREE)
+            ],
+            'net price configuration, net price is 129,79 => (30% off discount from 432,64)' => [
+                "9.08",
+                CartPrice::TAX_STATE_NET,
+                'EUR',
+                $this->createOrderLineItemEntity(1, 129.79, 129.79, 9.08, 7.0)
+            ],
```
boxblinkracer, line 100: please try this (MOL-70)
* expectedVatAmount = 1.89
* unitPrice = 1.1368
* taxRate = 16
* quantity = 12

boxblinkracer (changes requested):
please see comments. some optional but some definitely a good thing, also the unit test please

src/Service/ApiTaxCalculator.php
schmitzcarsten: the api tax calculator has been deleted and the algorithm has been implemented in orderservice directly. The rounding has been extracted to temporary variables for better readability

src/Service/OrderService.php
schmitzcarsten: the whole implementation has to be refactored in near future. This is only a fix for now

tests/Fake/LoggerFake.php
schmitzcarsten: refactored

tests/Service/OrderServiceTest.php
schmitzcarsten: both values $expected and $actual are strings. So same should be correct here, equals would except floats because of implicit cast

tests/Service/OrderServiceTest.php
schmitzcarsten: added

boxblinkracer (approved):

PR #76, opened 2021-02-16
MOL-159: do not send lifetimedays if they are not set in config

BlackScorp:
made the methods getOrderLifetimeDays and getOrderLifetimeDate nullable.

in the File Handler/Method/PaymentHandler.php:578-582

there is following line
```php 
         $dueDate = $settings->getOrderLifetimeDate();

            if ($dueDate !== null) {
                $orderData[self::FIELD_EXPIRES_AT] = $dueDate;
            }
```
so i assume if the $dueDate is not set it will not send to the api request. if the date is not configured in the plugin settings, this method will return null

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,20 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
-
 <phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/7.1/phpunit.xsd"
-         bootstrap="phpunit.autoload.php"
-         cacheResult="false">
-
-    <testsuites>
-        <testsuite name="MolliePayments Tests">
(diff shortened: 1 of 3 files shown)


phpunit.xml
```diff
@@ -1,20 +1,18 @@
...
+         colors="true"
+>
+  <coverage>
+    <include>
+      <directory>./src</directory>
+    </include>
+  </coverage>
+  <testsuites>
+    <testsuite name="MolliePayments Tests">
+      <directory>./oder kotests</directory>
```
barbieswimcrew, line 15: this directory doesn't look valid to me... could you please re-check that?

barbieswimcrew (changes requested):

phpunit.xml
BlackScorp: oh sorry, changed of course

barbieswimcrew (approved):

schmitzcarsten (approved):

PR #83, opened 2021-02-24
MOL-243: Empty cart after browser back

BlackScorp:
Listen to an event before going to Mollie page and store the last Order ID into the session, if the user decide to use browser back function, i am rebuilding the cart based on the orderId from the session.

this script will not work if user closes the browser since the session is dead at this point. we could just search of incomplete Orders and recreate a cart based of them, if required


barbieswimcrew (approved):

MelvinAchterhuis:
@schmitzcarsten , could you clarify why this PR was closed? Thank you!

PR #84, opened 2021-02-26
MOL-336: fix bank transfer dueDate fallback

schmitzcarsten:


barbieswimcrew (approved):

src/Setting/MollieSettingStruct.php
```diff
@@ -258,13 +263,22 @@ public function getPaymentMethodBankTransferDueDateDays(): int
...
+     * returns bank transfer due date in YYYY-MM-DD format or null
+     *
+     * @return string|null
      * @throws Exception
      */
-    public function getPaymentMethodBankTransferDueDate(): string
+    public function getPaymentMethodBankTransferDueDate(): ?string
     {
+        $dueDate = $this->getPaymentMethodBankTransferDueDateDays();
+        if (!$dueDate) {
```
boxblinkracer, line 274: in the code above the ! is ok i think
but here i would more rely on a null test to get into "real" checks instead of that "not value"?

boxblinkracer (approved):

PR #87, opened 2021-03-08
refactoring shipment api call

schmitzcarsten:
add shopware saas fixes

### DIFF
--- a/src/Facade/MollieShipment.php
+++ b/src/Facade/MollieShipment.php
@@ -71,7 +71,7 @@ class MollieShipment
 
         $customFields = $order->getCustomFields();
 
-        if (!isset($customFields[CustomFieldsInterface::MOLLIE_KEY][CustomFieldsInterface::ORDER_KEY])) {
+        if (!is_array($customFields) || !isset($customFields[CustomFieldsInterface::MOLLIE_KEY][CustomFieldsInterface::ORDER_KEY])) {
             $this->logger->warning(
                 sprintf('Mollie orderId does not exist in shopware order (%s)', (string)$order->getOrderNumber())
             );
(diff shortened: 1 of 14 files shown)


barbieswimcrew (approved):

src/Facade/MollieShipment.php
```diff
@@ -70,8 +70,9 @@ public function setShipment(string $orderDeliveryId, Context $context): bool
         }
 
         $customFields = $order->getCustomFields();
+        $mollieOrderId = $customFields[CustomFieldsInterface::MOLLIE_KEY][CustomFieldsInterface::ORDER_KEY] ?? null;
```
boxblinkracer, line 73: i think that might do a warning without isset, i would recommend adding that one again

src/Service/MolliePaymentExtractor.php
```diff
@@ -25,38 +27,31 @@ public function extractLast(?OrderTransactionCollection $collection): ?OrderTran
...
         // sort all transactions chronological
-        $collection->sort(function (OrderTransactionEntity $a, OrderTransactionEntity $b) {
+        $collection->sort(static function (OrderTransactionEntity $a, OrderTransactionEntity $b) {
             return $a->getCreatedAt() > $b->getCreatedAt();
         });
 
         $lastTransaction = $collection->last();
 
-        if ($this->isMolliePayment($lastTransaction)) {
+        if ($lastTransaction instanceof OrderTransactionEntity && $this->isMolliePayment($lastTransaction)) {
```
boxblinkracer, line 41: if there are mulitiple conditions, maybe early return?

boxblinkracer (approved):
looks good in general, please see commens, otherwise approved

src/Facade/MollieShipment.php
schmitzcarsten: no, this will not create a warning. It's like isset

PR #97, opened 2021-04-07
[MOL-368] fix apple pay

StephanAltmann85:


schmitzcarsten (approved):

schmitzcarsten (changes requested):
Is the pull request complete? You told me that there is existing a problem? I don't know if it's solved here?

StephanAltmann85:
> Is the pull request complete? You told me that there is existing a problem? I don't know if it's solved here?

The webhook controller will still change the shopware payment method to creditcard (if creditcard has been used with Apple Pay) afterwards. The fix would be to delete the update procedure in src/Helper/PaymentStatusHelper.php (Line 183). 

I don't understand why this has been implemented. For me it makes no sense at all, but I cannot make the decision for removing it.

schmitzcarsten (approved):

src/Helper/PaymentStatusHelper.php
```diff
@@ -156,7 +163,8 @@ public function processPaymentStatus(
 
         $useMollieFailureHandling = !$settings->isShopwareFailedPaymentMethod();
 
-        if ($useMollieFailureHandling) {
+        if ($useMollieFailureHandling
+            && $this->paymentMethodService->isPaidApplePayTransaction($transaction, $mollieOrder) === false) {
```
boxblinkracer, line 167: wouldnt it be correct to say "ifApplePay" transaction
otherwise it would still update it if its a failed apple pay method? then its suddenly credit card

so i guess the combination with "paid" is wrong?

src/MolliePayments.php
```diff
@@ -56,6 +57,12 @@ public function install(InstallContext $context) : void
     public function update(UpdateContext $context): void
     {
         parent::update($context);
+
+        if($context->getPlugin()->isActive() === true) {
+            // add domain verification
+            $domainVerificationService = $this->container->get(ApplePayDomainVerificationService::class);
```
boxblinkracer, line 63: why is it wrapped in an active flag?
what if the inactive plugin gets updated, then no download happens.
afterwards i would activate the plugin...does the postInstall from below get called?! if not, then the file is missing

src/Service/ApplePayDomainVerificationService.php
```diff
@@ -0,0 +1,29 @@
...
+    private const LOCAL_FILE = '/.well-known/apple-developer-merchantid-domain-association';
+
+    private $filesystem;
+
+    public function __construct(FilesystemInterface $filesystem) {
+        $this->filesystem = $filesystem;
+    }
+
+    public function downloadDomainAssociationFile() : void {
+        if($this->filesystem->has(self::LOCAL_FILE) === true) {
```
boxblinkracer, line 21: attention, if the signature of the file changes, it wouldnt be possible to update it without removing it.
i would rather always download the file

src/Helper/PaymentStatusHelper.php
StephanAltmann85: No, otherwise every ApplePay-Transaction would get excluded from MollieErrorHandling. A successfull ApplePayDirect transaction always should be paid.

src/MolliePayments.php
StephanAltmann85: If plugin isn't active the service is not available.

src/Service/ApplePayDomainVerificationService.php
StephanAltmann85: But it would also overwrite custom domain verification files.

src/Service/PaymentMethodService.php
```diff
@@ -334,9 +336,6 @@ public function getPaymentHandlers()
      */
     private function getMediaId(array $paymentMethod, Context $context): string
     {
-        /** @var string $mediaId */
```
boxblinkracer, line 337: what is going on in this file?
is that related to an apple pay fix?

src/Service/PaymentMethodService.php
```diff
@@ -347,33 +346,38 @@ private function getMediaId(array $paymentMethod, Context $context): string
...
+        );
+    }
+
+    /**
+     * @param OrderTransactionEntity $transaction
+     * @param Order $mollieOrder
+     * @return bool
+     */
+    public function isPaidApplePayTransaction(OrderTransactionEntity $transaction, Order $mollieOrder) : bool {
+        return $transaction->getPaymentMethod()->getHandlerIdentifier() === ApplePayPayment::class && $mollieOrder->isPaid() === true;
```
boxblinkracer, line 381: what about apple pay direct? is it covered with this in sw6?
(sw5 has 2 different payment methods)

StephanAltmann85: Only 1 payment method.

src/Service/PaymentMethodService.php
StephanAltmann85: No, just code clean up.

src/Service/ApplePayDomainVerificationService.php
```diff
@@ -0,0 +1,29 @@
...
+            return;
+        }
+
+        $content = file_get_contents(self::URL_FILE);
+
+        $this->filesystem->put(self::LOCAL_FILE, $content);
```
boxblinkracer, line 27: just to double check,...its available in the public docroot right?

src/Service/ApplePayDomainVerificationService.php
```diff
@@ -0,0 +1,29 @@
+<?php declare(strict_types=1);
+
+namespace Kiener\MolliePayments\Service;
+
+use League\Flysystem\FilesystemInterface;
+use Symfony\Component\DependencyInjection\ContainerAwareInterface;
+
+class ApplePayDomainVerificationService
+{
+    private const URL_FILE = 'https://www.mollie.com/.well-known/apple-developer-merchantid-domain-association';
```
boxblinkracer, line 10: please create unit tests to avoid accidental changes of these values, thx

boxblinkracer (changes requested):
see comments, thx

src/Service/ApplePayDomainVerificationService.php
StephanAltmann85: Yes

src/Service/ApplePayDomainVerificationService.php
StephanAltmann85: Done

src/Helper/PaymentStatusHelper.php
boxblinkracer: isnt this about avoiding that apple pay gets converted into credit card through a webhook?
yes every would excluded, but that is the aim as far as i know?!
@schmitzcarsten  please verify what you need

src/MolliePayments.php
boxblinkracer: yes but that was more the secondary question
what about the main problem

src/Service/ApplePayDomainVerificationService.php
boxblinkracer: custom domain files?!

tests/Service/ApplePayDomainVerificationServiceTest.php
```diff
@@ -0,0 +1,18 @@
...
+namespace Kiener\MolliePayments\Tests\Service;
+
+use Kiener\MolliePayments\Service\ApplePayDomainVerificationService;
+use PHPUnit\Framework\TestCase;
+
+class ApplePayDomainVerificationServiceTest extends TestCase
+{
+    public function testDomainVerificationContentRemainsUnchanged(): void
+    {
+        $content = file_get_contents(ApplePayDomainVerificationService::URL_FILE);
```
boxblinkracer, line 12: this test does not make any sense

please just verify the constant that the download URL is correct

tests/Fixtures/apple-developer-merchantid-domain-association
```diff
@@ -0,0 +1 @@
...33303131303630333535303430613063306134313730373036633635323034393665363332653331306233303039303630333535303430363133303235353533303230383638363066363939643963636137306633303064303630393630383634383031363530333034303230313035303061303831393533303138303630393261383634383836663730643031303930333331306230363039326138363438383666373064303130373031333031633036303932613836343838366637306430313039303533313066313730643331333933303335333133303331333033303339333133353561333032613036303932613836343838366637306430313039333433313164333031623330306430363039363038363438303136353033303430323031303530306131306130363038326138363438636533643034303330323330326630363039326138363438383666373064303130393034333132323034323035613437363366643264396534366338346162356331346462383563633833663831303934316536323838306363663138636536376131613630656633356661333030613036303832613836343863653364303430333032303434363330343430323230363436636338323861383361333062353136313731323266633462333532386432373762373937646264333861633064396263643439393864633832303634383032323030366663656534646432316661313165653665353834346561393565643465643034323939636666363333656437623233343461383835613433636431613662303030303030303030303030227D
```
boxblinkracer, line 1: please remove this

tests/Service/ApplePayDomainVerificationServiceTest.php
StephanAltmann85: Yes it makes sense, because it is testing that file url is valid and the file has not changed...

boxblinkracer: the file CAN change as mentioned above
and unit tests need to be able to run offline

tests/Fixtures/apple-developer-merchantid-domain-association
StephanAltmann85: Done

tests/Service/ApplePayDomainVerificationServiceTest.php
StephanAltmann85: done

src/MolliePayments.php
StephanAltmann85: File will be downloaded on activation.

boxblinkracer: ah perfect, github diff change made it look like its in postInstall :D all good thx

tests/Service/ApplePayDomainVerificationServiceTest.php
```diff
@@ -0,0 +1,16 @@
...
+use Kiener\MolliePayments\Service\ApplePayDomainVerificationService;
+use PHPUnit\Framework\TestCase;
+
+class ApplePayDomainVerificationServiceTest extends TestCase
+{
+    private const URL_FILE = 'https://www.mollie.com/.well-known/apple-developer-merchantid-domain-association';
+
+    public function testDomainVerificationUrlRemainsUnchanged(): void
+    {
+        self::assertEquals(ApplePayDomainVerificationService::URL_FILE, self::URL_FILE);
```
boxblinkracer, line 14: thank you, way better
please switch arguments, its (expected, actual)

src/MolliePayments.php
StephanAltmann85: Ich bin nach dem Kommentar auch darauf reingefallen.

tests/Service/ApplePayDomainVerificationServiceTest.php
StephanAltmann85: oh, ja

barbieswimcrew (approved):

schmitzcarsten (changes requested):

StephanAltmann85:
Done

schmitzcarsten (approved):

schmitzcarsten (approved):

PR #113, opened 2021-04-22
MOL-377: Fix wrong API key of specific channel in components

boxblinkracer:
if you have a specific configuration (api key) of a single sales channel, then this configuration has not been used.
it was always the MollieApiClient for the configuration for "All Sales Channels"

the origin of the problem was the singleton approach in the factory, i dont know where its created the first time, but it was with the wrong api key

i've tested: 
* before: var_dump with api key in components subscriber showed api key A (all) even though "storefront" had B
* after: var_dump shows api key B if enabled

i've improved the class itself a bit

IMPORTANT!
this factory fix could fix more problems related to refudns and shipping (where it has been used).
but it can also show problems "suddenly" when a merchant has a custom configuration that is wrong...at the moment its actually working...but wrong anyway :)

### DIFF
--- a/src/Subscriber/CheckoutConfirmPageSubscriber.php
+++ b/src/Subscriber/CheckoutConfirmPageSubscriber.php
@@ -3,6 +3,7 @@
 namespace Kiener\MolliePayments\Subscriber;
 
 use Exception;
+use Kiener\MolliePayments\Factory\MollieApiFactory;
 use Kiener\MolliePayments\Service\CustomerService;
 use Kiener\MolliePayments\Service\CustomFieldService;
 use Kiener\MolliePayments\Service\SettingsService;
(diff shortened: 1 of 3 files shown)


barbieswimcrew (approved):

src/Subscriber/CheckoutConfirmPageSubscriber.php
```diff
@@ -63,30 +80,39 @@ public static function getSubscribedEvents(): array
...
         $this->languageRepositoryInterface = $languageRepositoryInterface;
         $this->localeRepositoryInterface = $localeRepositoryInterface;
     }
 
     /**
-     * @param PageLoadedEvent|CheckoutConfirmPageLoadedEvent $args
+     * @param $args
+     * @throws \Mollie\Api\Exceptions\IncompatiblePlatform
      */
     public function addDataToPage($args): void
```
schmitzcarsten, line 101: $args is always instanceof CheckoutConfirmPageLoadedEvent ? 
If yes, please add to be type safe

schmitzcarsten: and add it to the private functions too please

schmitzcarsten (changes requested):

schmitzcarsten (approved):

PR #116, opened 2021-04-22
NTR: add php minimum check to pipeline

boxblinkracer:
add php minimum check for php 7.2 to plugin and pipelines

### DIFF
--- a/.github/workflows/ci_pipe.yml
+++ b/.github/workflows/ci_pipe.yml
@@ -10,8 +10,8 @@ on:
 
 jobs:
 
-  syntax_checks:
-    name: Syntax Checks | PHP ${{ matrix.php }}
+  php_checks:
+    name: PHP Syntax Checks | PHP ${{ matrix.php }}
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -66,5 +66,9 @@
     "phpstan/phpstan": "^0.12.82",
     "friendsofphp/php-cs-fixer": "^2.18",
     "nunomaduro/phpinsights": "^1.14"
+  },
+  "scripts": {
```
boxblinkracer, line 70: here it is 
maybe wew should add a ignore or | true to it for production compatibility?

barbieswimcrew (approved):

composer.json
boxblinkracer: solved

.github/workflows/ci_pipe.yml
```diff
@@ -26,13 +26,32 @@ jobs:
...
+    needs: php_checks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Code
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.4
```
schmitzcarsten, line 42: wrong php version? Should be 7.2 or am i missing the point here?

.github/workflows/pr_pipe.yml
```diff
@@ -23,13 +23,32 @@ jobs:
...
+    needs: php_checks
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Code
+        uses: actions/checkout@v2
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.4
```
schmitzcarsten, line 39: should be 7.2?

schmitzcarsten (changes requested):

.github/workflows/pr_pipe.yml
boxblinkracer: nope, this is just for dev dependency installation
the check internally is scripted in the makefile and points to php 7.2

PR #120, opened 2021-04-29
MOL-359: add 6.4 compatibility

schmitzcarsten:
added administration routes without versioning

updated test button in administration

added solution for new rounding feature


src/Controller/Api/ConfigController.php
```diff
@@ -32,6 +30,31 @@ public function testApiKeys(Request $request): JsonResponse
...
+        // Get the live API key
+        $liveApiKey = $request->get('liveApiKey');
+
+        // Get the test API key
+        $testApiKey = $request->get('testApiKey');
+
+        return $this->getResponse($liveApiKey, $testApiKey);
+    }
+
+    private function getResponse(string $liveApiKey, string $testApiKey): JsonResponse
```
barbieswimcrew, line 56: instead of building the JsonResponse's dataset structure inside a private method I'd prefer to create a specific response class... for example something like...

`ApiKeyResponse extends JsonResponse implements ResponseInterface`

The benefit would be that you'd easily test that by writing unit tests against that

src/Controller/Api/RefundController.php
```diff
@@ -275,6 +296,31 @@ public function refund(Request $request): JsonResponse
...
+     */
+    public function total64(Request $request): JsonResponse
+    {
+        /** @var string|null $orderId */
+        $orderId = $request->get('orderId');
+
+        return $this->getTotalResponse($orderId);
+    }
+
+    private function getTotalResponse(?string $orderId): JsonResponse
```
barbieswimcrew, line 323: same like above... something like

`TotalResponse extends JsonResponse implements ResponseInterface`

would do the trick

src/Controller/Api/ShippingController.php
```diff
@@ -231,9 +231,208 @@ public function ship(Request $request): JsonResponse
...
+     * @Route("/api/v{version}/_action/mollie/ship",
+     *         defaults={"auth_enabled"=true}, name="api.action.pre64.mollie.ship", methods={"POST"})
+     *
+     * @param Request $request
+     *
+     * @return JsonResponse
+     * @throws ApiException
+     * @throws IncompatiblePlatform
+     */
+    public function oldShip(Request $request): JsonResponse
```
barbieswimcrew, line 246: this method looks way to complex with all that conditions... 'd love to see that refactored

barbieswimcrew (changes requested):

.github/workflows/pr_pipe.yml
```diff
@@ -107,59 +107,3 @@ jobs:
...
-
-      - name: Install Dependencies
-        run: make dev -B
-
-      - name: Run PHPInsights Analyser
-        run: make insights -B
-
-  # ------------------------------------------------------------------------------------------------------------------------
-
-  integration:
```
boxblinkracer, line 133: why is this removed?
it would show if the plugin can be installed

src/Controller/Api/RefundController.php
```diff
@@ -99,6 +99,27 @@ public function __construct(
...
+     * @Route("/api/mollie/refund",
+     *         defaults={"auth_enabled"=true}, name="api.action.mollie.refund-64", methods={"POST"})
+     *
+     * @param Request $request
+     *
+     * @return JsonResponse
+     * @throws ApiException
+     * @throws IncompatiblePlatform
+     */
+    public function refund64(Request $request): JsonResponse
```
boxblinkracer, line 117: i guess this is a shopware 6.4 route? maybe add a comment? :D i thought its some kind of base64 refunding

src/Controller/Api/ShippingController.php
```diff
@@ -231,9 +231,208 @@ public function ship(Request $request): JsonResponse
...
+            (string)$orderId !== ''
+            && (string)$orderLineId !== ''
+            && $quantity > 0
+        ) {
+            $apiClient = $this->apiFactory->createClient(
+                $orderLineItem->getOrder()->getSalesChannelId()
+            );
+        }
+
+        if ($apiClient !== null) {
```
boxblinkracer, line 317: i dont know if this is old and moved code?
but i would rather see more early returns if its a new code

src/Controller/Api/ShippingController.php
```diff
@@ -231,9 +231,208 @@ public function ship(Request $request): JsonResponse
...
+            $settings = $this->settingsService->getSettings(
+                $orderLineItem->getOrder()->getSalesChannelId()
+            );
+
+            /** @var array $orderParameters */
+            $orderParameters = [];
+
+            if ($settings->isTestMode() && $apiClient->usesOAuth()) {
+                $orderParameters = [
+                    self::SHIPPING_DATA_KEY_TEST_MODE => true
```
boxblinkracer, line 328: only test mode with oauth? not without?

src/Controller/Api/ShippingController.php
```diff
@@ -231,9 +231,208 @@ public function ship(Request $request): JsonResponse
...
+                    self::SHIPPING_DATA_KEY_LINES => [
+                        [
+                            self::SHIPPING_DATA_KEY_ID => $orderLineId,
+                            self::SHIPPING_DATA_KEY_QUANTITY => $quantity,
+                        ],
+                    ],
+                ];
+
+                if ($settings->isTestMode() && $apiClient->usesOAuth()) {
+                    $shipmentData[self::SHIPPING_DATA_KEY_TEST_MODE] = true;
```
boxblinkracer, line 345: only test mode with oauth? not without?

src/Handler/PaymentHandler.php
```diff
@@ -228,9 +228,20 @@ public function pay(
                 );
             }
 
-            // Get the payment url from the order at Mollie.
-            if ($mollieOrder !== null) {
-                $paymentUrl = isset($mollieOrder) ? $mollieOrder->getCheckoutUrl() : null;
+            if (!$mollieOrder instanceof Order) {
+                throw new Exception('Couldn\'t create payment at mollie');
```
boxblinkracer, line 232: just a small one, if you use " then you wouldnt need \'

src/Handler/PaymentHandler.php
```diff
@@ -228,9 +228,20 @@ public function pay(
...
+            }
+
+            // first check if we got a checkoutUrl from mollie (in case of creditcard components
+            // it could be that payment is already finished, checkout url misses in these cases)
+            $paymentUrl = $mollieOrder->getCheckoutUrl();
+
+            if ($paymentUrl === null) {
+                if ($mollieOrder->isPaid()) {
```
boxblinkracer, line 240: NO!?!
attention, this breaks payments that are authorized, or open....so all positive payments that are not "paid".
could that be?

src/Service/OrderService.php
```diff
@@ -188,11 +203,83 @@ public function getOrderLinesArray(OrderEntity $order): array
...
+        // Build the order lines array
+        return [
+            'type' => OrderLineType::TYPE_DISCOUNT,
+            'name' => $itemName,
+            'quantity' => 1,
+            'unitPrice' => $this->getPriceArray($currencyCode, $roundedLineItemTotalPrice),
+            'totalAmount' => $this->getPriceArray($currencyCode, $roundedLineItemTotalPrice),
+            'vatRate' => number_format(0.0, self::MOLLIE_PRICE_PRECISION, '.', ''),
+            'vatAmount' => $this->getPriceArray($currencyCode, 0.00),
+            'sku' => 'SHOPWARE ROUNDING ITEM'
```
boxblinkracer, line 279: small thing, but maybe avoid blanks in sku?!

src/Validator/OrderTotalRoundingValidator.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * function validates to true if we have to consider new 6.4 rounding feature
+     *
+     * @param int $decimals
+     * @param bool $roundForNet
+     * @param float $interval
+     * @param string $orderTaxState
+     * @return bool
+     */
+    public function validate(int $decimals, bool $roundForNet, float $interval, string $orderTaxState): bool
```
boxblinkracer, line 18: wouldnt it be better to rename it to something like "shouldRound"....validate might be misleading

src/Validator/OrderTotalRoundingValidator.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * function tells us if we could consider that rounding configuration would lead to new
+     * rounding behaviour
+     *
+     * @param int $decimals
+     * @param bool $roundForNet
+     * @param string $taxState
+     * @return bool
+     */
+    public function validateShouldRound(int $decimals, bool $roundForNet, string $taxState): bool
```
boxblinkracer, line 37: ok i dont get the difference between these 2 functions
they are both "should consider new rounding of 6.4"
but what is the difference? i get that the second function is more about conditions than the actual value, but 2 public functions that do almos the same and almost have the same description are not so good.
option to make it private?

tests/PHPUnit/Validator/OrderTotalRoundingValidatorTest.php
```diff
@@ -0,0 +1,86 @@
...
+        $this->assertFalse((new OrderTotalRoundingValidator())->validate($decimals, $roundForNet, $interval, $orderTaxState));
+    }
+
+    /**
+     * wrong tax state config => validator should return true
+     *
+     * if the tax state is configured that net rounding should be available and we have a NET order,
+     * then rounding could be responsible for price differences
+     */
+    public function testValidateCorrectTaxState(): void
```
boxblinkracer, line 61: had problems differ these 2 tests
maybe name them to something like "testValidateCorrectTaxStateNetRounding" and the upper one GrossRounding?

src/Service/OrderService.php
```diff
@@ -188,11 +203,83 @@ public function getOrderLinesArray(OrderEntity $order): array
...
+     * The function will only return a lineItem if
+     * - shopware version >= 6.4
+     * - total amount rounding is active and configured that differences could exist
+     * - and if discrepances do exist at all
+     *
+     * @param OrderEntity $order
+     * @param string $currencyCode
+     * @return array
+     */
+    public function getRoundingLineItem(OrderEntity $order, string $currencyCode): array
```
boxblinkracer, line 235: this would be a perfect unit test, i think i didnt see one below for this?

src/Controller/Api/ConfigController.php
schmitzcarsten: I agree with you. But the problem with the admin api controllers has been that the versioning of the route has been removed. So I have to implement another route without the versioning and keep the old because of backwards compatibility reasons. I just extracted the logic into private method and use them in both routes.

No time to refactor this at the moment and it is old code

src/Controller/Api/ShippingController.php
schmitzcarsten: it is old code, see my comments to @barbieswimcrew comment for the reasons why old code became "new" code

src/Controller/Api/ShippingController.php
schmitzcarsten: old code, needs refactoring in future

src/Controller/Api/ShippingController.php
schmitzcarsten: old code

src/Handler/PaymentHandler.php
schmitzcarsten: updated like discussed today

tests/PHPUnit/Validator/OrderTotalRoundingValidatorTest.php
schmitzcarsten: update function to private

src/Validator/OrderTotalRoundingValidator.php
schmitzcarsten: update function to private

src/Service/OrderService.php
schmitzcarsten: will be part of future refactoring. We will add some kind of a LineItemBuilder then. unit test will follow

.github/workflows/ci_pipe.yml
```diff
@@ -166,63 +146,3 @@ jobs:
         run: |
           docker exec shop bash -c "php bin/console plugin:refresh"
           docker exec shop bash -c "php bin/console plugin:install MolliePayments --activate"
-
-  # ------------------------------------------------------------------------------------------------------------------------
-
-  e2e:
```
boxblinkracer, line 172: this would remove cypress

schmitzcarsten: added and resolved merge conflict....

PR #132, opened 2021-05-21
Fix ideal issuers select in checkout

MelvinAchterhuis:
This PR fixes the iDeal issuer dropdown in the checkout in 6.4

### DIFF
--- a/src/Resources/app/storefront/src/mollie-payments/plugins/ideal-issuer.plugin.js
+++ b/src/Resources/app/storefront/src/mollie-payments/plugins/ideal-issuer.plugin.js
@@ -19,10 +19,23 @@ export default class MollieIDealIssuer extends Plugin {
 
             // Elements
             const customerId = container.getAttribute('data-customer-id');
-            const paymentForm = document.querySelector('#confirmPaymentForm');
-            const submitButton = document.querySelector('#confirmPaymentForm button[type="submit"]');
-            const radioInputs = document.querySelectorAll('#confirmPaymentForm input[type="radio"]');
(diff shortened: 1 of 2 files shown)


src/Resources/app/storefront/src/mollie-payments/plugins/ideal-issuer.plugin.js
```diff
@@ -19,10 +19,23 @@ export default class MollieIDealIssuer extends Plugin {
...
-            const submitButton = document.querySelector('#confirmPaymentForm button[type="submit"]');
-            const radioInputs = document.querySelectorAll('#confirmPaymentForm input[type="radio"]');
-            const iDealRadioInput = document.querySelector('#confirmPaymentForm input[type="radio"].ideal');
+            let paymentForm;
+            let submitButton;
+            let radioInputs;
+            let iDealRadioInput;
+
+            // Form id changed between 6.3 and 6.4
+            if (document.querySelector('#confirmPaymentForm')) {
```
barbieswimcrew, line 28: to reduce the lines of code i'd recommend to do like this:

```
let confirmPaymentForm = document.querySelector('#confirmPaymentForm');
let changePaymentForm = document.querySelector('#changePaymentForm');

if(confirmPaymentForm){
    paymentForm = confirmPaymentForm;
} else if (checkoutPaymentForm) {
    paymentForm = checkoutPaymentForm;
}

submitButton = paymentForm.querySelector('button[type="submit"]');
// ...and so on...
```

barbieswimcrew (changes requested):

schmitzcarsten:
closed because of inactivity. We will add compatibility in another PR

PR #137, opened 2021-06-03
MOL-209 - Add a link to the Mollie payment screen to orders (backend)

KienerNL:
Added this as a button that copies the link to the user's clipboard. Button is disabled when link is unavailable, for example, when the order has been paid already. Button is not shown for non Mollie orders.


src/Service/MollieApi/Order.php
```diff
@@ -26,6 +26,27 @@ public function __construct(MollieApiFactory $clientFactory, LoggerInterface $lo
         $this->logger = $logger;
     }
 
+    public function getPaymentUrl(string $mollieOrderId, string $salesChannelId): ?string
+    {
+        $apiClient = $this->clientFactory->getClient($salesChannelId);
```
schmitzcarsten, line 31: duplicate code from line 31 to 48 (coming from setShipment)

extract this to an own public function getMollieOrder and use this then  in both functions

schmitzcarsten (changes requested):
If an order is paid or refunded you deactivate the copy link (makes sense since not necessary because order is paid)

But in case the order has been cancelled you may copy the link, but it does not work (as order is cancelled at mollie). So we have to deactivate the copy function in this cases. Please check state failed (should be deactivated too)

src/Service/MollieApi/Order.php
KienerNL: Moved to a separate function

schmitzcarsten (approved):

src/Resources/app/administration/src/module/mollie-payments/components/mollie-test-api-key/mollie-test-api-key.html.twig
```diff
@@ -1,3 +1,8 @@
-<sw-button @click="onTestButtonClicked" ref="testApiButton">
-    {{ $tc('sw-payment.testButton') }}
-</sw-button>
+<sw-container gap="16px" columns="1fr" justify="start">
+    <sw-button @click="onTestButtonClicked" ref="testApiButton">
+        {{ $tc('sw-payment.testButton') }}
+    </sw-button>
+    <sw-button variant="context" size="x-small" link="https://www.mollie.com/dashboard/developers/api-keys">
```
boxblinkracer, line 5: isnt this another feature? deeplink to the api keys page?

src/Service/MollieApi/Order.php
```diff
@@ -43,6 +43,18 @@ public function setShipment(string $mollieOrderId, string $salesChannelId): bool
 
             throw $e;
         }
+    }
+
+    public function getPaymentUrl(string $mollieOrderId, string $salesChannelId): ?string
+    {
+        $mollieOrder = $this->getMollieOrder($mollieOrderId, $salesChannelId);
```
boxblinkracer, line 50: wouldnt that be a great spot to get rid off null and use an empty string?

also, i did the featue in sw5, and as far as i can remember, once the payment or order isnt valid anymore (because it was paid) then the payment url is null anyway
so i guess the disabling doesnt make sense either...its just empty, in case it helps.
i did a placeholder instead showing "No Checkout URL available"

src/Resources/app/administration/src/module/mollie-payments/components/mollie-test-api-key/mollie-test-api-key.html.twig
KienerNL: Yes, which has already been merged into master. And then this branch was made up to date with master to prevent any conflicts.

boxblinkracer (approved):

PR #144, opened 2021-06-11
MOL-199 Partial Refunds

KienerNL:
PR #89 and #134


src/Service/RefundService.php
```diff
@@ -0,0 +1,236 @@
...
+    {
+        return $this->getPaymentForOrder($order)->getAmountRefunded();
+    }
+
+    /**
+     * @param OrderEntity $order
+     * @return Payment
+     * @throws MollieRefundException
+     */
+    private function getPaymentForOrder(OrderEntity $order): Payment
```
schmitzcarsten, line 176: extract this private function to own service. Use as parameters mollieOrderId (if empty, don't do lookup at all) and the saleschannelcontextid from order. This will make unit test easier as parameters are simple parameter types

src/Service/RefundService.php
```diff
@@ -0,0 +1,236 @@
...
+                sprintf("Could not fetch refunds for order %s (Order number %s)",
+                    $payment->orderId,
+                    $order->getOrderNumber()
+                )
+            );
+        }
+
+        // Apparently Refund::amount and Refund::settlementAmount don't json encode very well, resulting in an empty
+        // array, so we build an array manually.
+        return array_map(function ($refund) {
```
schmitzcarsten, line 121: what about a hydratorhere, that builds an array out of a refund. You could unit test this hydrator as separate unit test. Here everything would be a foreach and inside using hydrator

src/Service/RefundService.php
```diff
@@ -0,0 +1,236 @@
...
+                'id' => $refund->id,
+                'orderId' => $refund->orderId,
+                'paymentId' => $refund->paymentId,
+                'amount' => !is_null($refund->amount)
+                    ? [
+                        'value' => $refund->amount->value,
+                        'currency' => $refund->amount->currency,
+                    ]
+                    : null,
+                'settlementAmount' => !is_null($refund->settlementAmount)
```
schmitzcarsten, line 133: this is not very readable here. Why not

$array=[something];
if (!is_null($refund->settlementAmount)) {
    $array['settlementAmount']=[something]
}

src/Service/RefundService.php
```diff
@@ -0,0 +1,236 @@
...
+        $apiClient = $this->getApiClientForOrder($order);
+
+        if (!($apiClient instanceof MollieApiClient)) {
+            throw new MollieRefundException("Could not create a Mollie Api Client");
+        }
+
+        $mollieOrderId = $order->getCustomFields()['mollie_payments']['order_id'] ?? '';
+
+        if (empty($mollieOrderId)) {
+            throw new MollieRefundException(
```
schmitzcarsten, line 187: better is to have explicit Exceptions here 
perhaps CouldNotExtractMollieOrderId, then CouldNotFetchMollieOrder, then RefundsNotFound

src/Subscriber/PaymentStateSubscriber.php
```diff
@@ -58,7 +58,7 @@ public function onOrderTransactionChanged(StateMachineStateChangeEvent $event)
 
         try {
             $this->setMollieOrderRefunded->setRefunded($event->getTransition()->getEntityId(), $event->getContext());
-        } catch (CouldNotSetRefundAtMollieException $e) {
+        } catch (MollieRefundException $e) {
```
schmitzcarsten, line 61: find a better exception name for the usecase

tests/PHPUnit/Facade/SetMollieOrderRefundedTest.php
```diff
@@ -57,80 +53,121 @@ public function testThrowExceptionIfTransactionCouldNotBeFound(): void
     {
         $this->transactionService->method('getTransactionById')->willReturn(null);
 
-        self::expectException(CouldNotSetRefundAtMollieException::class);
+        self::expectException(MollieRefundException::class);
```
schmitzcarsten, line 56: without explicit exceptions you don't know if error is thrown in the top or the bottom. So here the unit test could be green but the error could occure at an other line of code that should not be in the  scope of this test

tests/PHPUnit/Service/MollieRefundsServiceTest.php
```diff
@@ -0,0 +1,506 @@
...
+use Mollie\Api\Resources\PaymentCollection;
+use Mollie\Api\Resources\Refund;
+use Mollie\Api\Resources\RefundCollection;
+use Mollie\Api\Types\PaymentStatus;
+use Mollie\Api\Types\RefundStatus;
+use PHPUnit\Framework\TestCase;
+use Shopware\Core\Checkout\Order\OrderEntity;
+use Shopware\Core\System\Currency\CurrencyEntity;
+
+class MollieRefundsServiceTest extends TestCase
```
schmitzcarsten, line 22: the class that you want to test is RefundService, so the correct unit test class name is RefundServiceTest. Perhaps  someone implements an MollieRefundsService in future ;)

schmitzcarsten (changes requested):

.phpstan.lvl8.neon
```diff
@@ -6,6 +6,7 @@ parameters:
         - ./src/Controller/*
         - ./src/Handler/*
         - ./src/Helper/*
+        - ./src/Hydrator/*
```
boxblinkracer, line 9: hi
if i got this correctly, then the hydrator folder is new?...means that all files in there are new as well?
then please ensure that your new files meet the phpstan level requirements instead of ignoring it

src/Facade/SetMollieOrderRefunded.php
```diff
@@ -57,35 +51,20 @@ public function setRefunded(string $orderTransactionId, Context $context): void
             );
         }
 
-        $customFields = $order->getCustomFields() ?? [];
+        $refunded = $this->refundService->getRefundedAmount($order);
+        $toRefund = $order->getAmountTotal() - $refunded;
```
boxblinkracer, line 55: im not quite sure, but i once saw a "amountRemaining" property in a mollie object...i think this is fine here, just wanted to let you know this

src/Service/MollieApi/Order.php
```diff
@@ -67,4 +80,34 @@ public function setShipment(string $mollieOrderId, string $salesChannelId): bool
...
+     * @param string $mollieOrderId
+     * @param string $salesChannelId
+     * @return Payment
+     */
+    public function getCompletedPayment(string $mollieOrderId, string $salesChannelId): Payment
+    {
+        try {
+            $mollieOrder = $this->getMollieOrder($mollieOrderId, $salesChannelId, ['embed' => 'payments']);
+        } catch (ApiException $e) {
+            throw new CouldNotFetchMollieOrderException($mollieOrderId);
```
boxblinkracer, line 94: could you append the previous exception or add a log or anything
we never know the cause of the exception otherwise

boxblinkracer: or just remove the try catch if there should always be an order? then just fail-fast automatically?

src/Service/MollieApi/Order.php
```diff
@@ -67,4 +80,34 @@ public function setShipment(string $mollieOrderId, string $salesChannelId): bool
...
+        }
+
+        if ($mollieOrder->payments()->count() === 0) {
+            throw new PaymentNotFoundException($mollieOrderId);
+        }
+
+        /** @var Payment $payment */
+        foreach ($mollieOrder->payments()->getArrayCopy() as $payment) {
+            if (in_array($payment->status, [
+                PaymentStatus::STATUS_PAID,
```
boxblinkracer, line 104: doesnt klarna with orders api also have a COMPLETED state? so a different one from paid?

src/Service/RefundService.php
```diff
@@ -0,0 +1,241 @@
...
+            );
+
+            throw $e;
+        }
+
+        try {
+            // getRefund doesn't contain all necessary @throws tags.
+            // It is possible for it to throw an ApiException here if $refundId is incorrect.
+            $refund = $payment->getRefund($refundId);
+        } catch (ApiException $e) { // Invalid resource id
```
boxblinkracer, line 141: i think we need to log that error right?

src/Service/RefundService.php
```diff
@@ -0,0 +1,241 @@
...
+        // Refunds can only be cancelled when they're still queued or pending.
+        if (!$refund->isQueued() && !$refund->isPending()) {
+            return false;
+        }
+
+        try {
+            $refund->cancel();
+            return true;
+        } catch (ApiException $e) {
+            throw new CouldNotCancelMollieRefundException($mollieOrderId, $order->getOrderNumber(), $refundId);
```
boxblinkracer, line 159: i think we need to log that too?

schmitzcarsten (changes requested):

src/Service/RefundService.php
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * @param OrderEntity $order
+     * @return array
+     */
+    public function getRefunds(OrderEntity $order): array
+    {
+        try {
+            $mollieOrderId = $this->orderService->getMollieOrderId($order);
+        } catch (CouldNotExtractMollieOrderIdException $e) {
+            // This may not be a Mollie order.
```
boxblinkracer, line 172: wouldnt that be more an exception because something is wrong?
if nothing is logged or thrown then this is a typical scenario with a fail-safe approach..but from the comment it seems as fi something might be broken? because of a missing id?

src/Service/RefundService.php
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * @param OrderEntity $order
+     * @return float
+     */
+    public function getRemainingAmount(OrderEntity $order): float
+    {
+        try {
+            $mollieOrderId = $this->orderService->getMollieOrderId($order);
+        } catch (CouldNotExtractMollieOrderIdException $e) {
+            // This may not be a Mollie order.
```
boxblinkracer, line 205: also here, it seems as if this is no expected behaviour? then a log or exception would be nice

src/Service/RefundService.php
```diff
@@ -0,0 +1,241 @@
...
+    /**
+     * @param OrderEntity $order
+     * @return float
+     */
+    public function getRefundedAmount(OrderEntity $order): float
+    {
+        try {
+            $mollieOrderId = $this->orderService->getMollieOrderId($order);
+        } catch (CouldNotExtractMollieOrderIdException $e) {
+            // This may not be a Mollie order.
```
boxblinkracer, line 228: also here, it seems as if this is no expected behaviour? then a log or exception would be nice

tests/PHPUnit/Facade/SetMollieOrderRefundedTest.php
```diff
@@ -69,68 +66,107 @@ public function testThrowExceptionIfTransactionHasNoOrder(): void
...
-        $order = $this->getOrder(Uuid::randomHex(), Uuid::randomHex());
-        $this->transactionService->method('getTransactionById')->willReturn($this->getTransaction(Uuid::randomHex(), $order));
-
-        self::expectException(CouldNotSetRefundAtMollieException::class);
-        $this->setMollieOrderService->setRefunded('foo', $this->context);
-    }
-
-    public function testThatApiClientIsConstructedWithCorrectSalesChannel(): void
+    /**
+     * Test refunds
```
boxblinkracer, line 70: im happy that you provide a comment here. i do the same with every unit test.
but could you be a bit more specific on what is done in here, what is expected and things like that. because that comment doesnt have any real benefit otherwise :)

tests/PHPUnit/Service/RefundsServiceTest.php
```diff
@@ -0,0 +1,511 @@
...
+
+        $this->clientMock = $this->createMock(MollieApiClient::class);
+
+        $this->orderService = new OrderService(
+            $this->createMock(EntityRepositoryInterface::class),
+            $this->createMock(EntityRepositoryInterface::class),
+            $logger,
+            new OrderLineItemValidator($logger),
+            new OrderTotalRoundingValidator(),
+            '6.3.5.4'
```
boxblinkracer, line 49: does this test only cover a specific shopware version?
or what is that number and what causes it?

boxblinkracer (changes requested):
i did a few comments

tests/PHPUnit/Service/RefundsServiceTest.php
KienerNL: The OrderService normally gets the `%kernel.shopware_version%` passed into in its constructor. In this test a fixed value should suffice. Also this was copied from the OrderServiceTest where @schmitzcarsten did the same.

tests/PHPUnit/Facade/SetMollieOrderRefundedTest.php
```diff
@@ -69,68 +66,107 @@ public function testThrowExceptionIfTransactionHasNoOrder(): void
...
+     *
+     * @param float $amountTotal
+     * @param float $amountRefunded
+     * @param float|null $expectedRefund
+     * @dataProvider getRefundsTestData
+     */
+    public function atestThatRefundIsDone(
```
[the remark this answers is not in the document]
KienerNL: Oops :see_no_evil:

tests/PHPUnit/Service/RefundsServiceTest.php
boxblinkracer: ok thx
i was asking because there is a different behaviour in the order service with shopware >= 6.4 which affects rounded items.
if that would affect refunds (which i dont think so) then you would need to do an additional 6.4-variant of your tests
otherwise feel free to mark this comment as resolved :)

src/Service/RefundService.php
KienerNL: I was going to write an explanation why, but then I realized we probably should not even call these methods if they are not Mollie orders to begin with. So I added a check in the `sw-order-detail-base` view to check if there's a Mollie order id available, before calling these methods. These methods now log and re-throw the error if the order id cannot be found.

src/Service/RefundService.php
KienerNL: See above.

src/Service/RefundService.php
KienerNL: See above.

tests/PHPUnit/Facade/SetMollieOrderRefundedTest.php
KienerNL: Comment has been updated.

src/Service/RefundService.php
```diff
@@ -0,0 +1,241 @@
...
+        } catch (PaymentNotFoundException $e) {
+            // This mollie order may not be paid yet, so there can't be any refunds.
+            return [];
+        }
+
+        try {
+            $refundsArray = [];
+
+            foreach ($payment->refunds()->getArrayCopy() as $refund) {
+                $refundsArray[] = RefundHydrator::hydrate($refund);
```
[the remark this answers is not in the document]
KienerNL: Static removed.

src/Service/RefundService.php
boxblinkracer: perfect thx :)

src/Hydrator/RefundHydrator.php
```diff
@@ -0,0 +1,47 @@
...
+
+use Mollie\Api\Resources\Refund;
+
+class RefundHydrator
+{
+    /**
+     * @param Refund $refund
+     * @return array
+     */
+    public static function hydrate(Refund $refund): array
```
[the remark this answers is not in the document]
KienerNL: See above.

.phpstan.lvl8.neon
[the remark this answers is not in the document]
KienerNL: I've sent you a message on Slack about the errors I'm getting here

src/Facade/SetMollieOrderRefunded.php
KienerNL: Yes, but also no. It has come to my attention that some accounts are able to refund an extra amount on top of the order total. So if the order total is € 100 they can refund up to € 125. The Kiener account we use to test is able to do this, but a different account cannot. I'll be asking Lono about this in our next meet.

If the account can refund more than the order total, then this is included in the `amountRemaining` property you mentioned. Therefor I'd rather use the order total from the Shopware order to calculate how much needs to be refunded. If the customer can refund more and wants to refund more, then it would be better to use a partial refund to refund everything.

.phpstan.lvl8.neon
KienerNL: This is now solved with @schmitzcarsten 's help

tests/PHPUnit/Service/RefundsServiceTest.php
KienerNL: Fortunately the refund service only uses the `getMollieOrderId` from the order service. ;)

src/Service/MollieApi/Order.php
schmitzcarsten: Authorized is not correct here. It should be paid or completed. see 
Kiener\MolliePayments\Service\Mollie\MolliePaymentStatus::MOLLIE_PAYMENT_COMPLETED

src/Service/RefundService.php
```diff
@@ -0,0 +1,202 @@
...
+
+    /**
+     * @param OrderEntity $order
+     * @return string
+     * @throws CouldNotExtractMollieOrderIdException
+     */
+    private function tryGetMollieOrderId(OrderEntity $order): string
+    {
+        try {
+            return $this->orderService->getMollieOrderId($order);
```
schmitzcarsten, line 196: remove function and use $this->orderService->getMollieOrderId direct allover this class. 

You are logging multiple errors outside in the controller

src/Controller/Api/RefundController.php
```diff
@@ -63,326 +39,266 @@ class RefundController extends AbstractController
...
+    private function refundResponse(string $orderId, float $amount, Context $context): JsonResponse
+    {
+        try {
+            $order = $this->getValidOrder($orderId, $context);
+
+            $success = $this->refundService->refund($order, $amount);
+        } catch (ShopwareHttpException $e) {
+            $this->logger->error($e->getMessage());
+            return $this->json(['message' => $e->getMessage()], $e->getStatusCode());
+        } catch (CouldNotCreateMollieRefundException | CouldNotExtractMollieOrderIdException | CouldNotFetchMollieOrderException | PaymentNotFoundException $e) {
```
schmitzcarsten, line 196: all errors (except ShopwareHttpException catch) are doing the same here. better catch throwable here, log it and return the json. So you would catch all errors, log them and return here

src/Controller/Api/RefundController.php
```diff
@@ -63,326 +39,266 @@ class RefundController extends AbstractController
...
+    private function cancelResponse(string $orderId, ?string $refundId, Context $context): JsonResponse
+    {
+        try {
+            $order = $this->getValidOrder($orderId, $context);
+
+            $success = $this->refundService->cancel($order, $refundId);
+        } catch (ShopwareHttpException $e) {
+            $this->logger->error($e->getMessage());
+            return $this->json(['message' => $e->getMessage()], $e->getStatusCode());
+        } catch (CouldNotCancelMollieRefundException | CouldNotExtractMollieOrderIdException | CouldNotFetchMollieOrderException | PaymentNotFoundException $e) {
```
schmitzcarsten, line 221: catch throwable here

src/Controller/Api/RefundController.php
```diff
@@ -63,326 +39,266 @@ class RefundController extends AbstractController
...
+    private function listResponse(string $orderId, Context $context): JsonResponse
+    {
+        try {
+            $order = $this->getValidOrder($orderId, $context);
+
+            $refunds = $this->refundService->getRefunds($order);
+        } catch (ShopwareHttpException $e) {
+            $this->logger->error($e->getMessage());
+            return $this->json(['message' => $e->getMessage()], $e->getStatusCode());
+        } catch (CouldNotExtractMollieOrderIdException | CouldNotFetchMollieOrderException | CouldNotFetchMollieRefundsException $e) {
```
schmitzcarsten, line 245: move this catch under paymentNotFound and catch throwable here

src/Controller/Api/RefundController.php
```diff
@@ -63,326 +39,266 @@ class RefundController extends AbstractController
...
-        $orderLineCriteria = new Criteria([$lineItemId]);
+        try {
+            $order = $this->getValidOrder($orderId, $context);
+
+            $remaining = $this->refundService->getRemainingAmount($order);
+            $refunded = $this->refundService->getRefundedAmount($order);
+        } catch (ShopwareHttpException $e) {
+            $this->logger->error($e->getMessage());
+            return $this->json(['message' => $e->getMessage()], $e->getStatusCode());
+        } catch (CouldNotExtractMollieOrderIdException | CouldNotFetchMollieOrderException $e) {
```
schmitzcarsten, line 271: same here catch throwable and move the catch under paymentnotfoundexception

tests/PHPUnit/Hydrator/RefundHydratorTest.php
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private function createRefundTestData(?float $amount, ?float $settlementAmount): array
+    {
+        $id = 're_' . Random::getAlphanumericString(10);
+        $orderId = 'ord_' . Random::getAlphanumericString(6);
+        $paymentId = 'tr_' . Random::getAlphanumericString(10);
+        $description = Random::getAlphanumericString(100);
+        $createdAt = date(DATE_ISO8601);
+        $status = Random::getRandomArrayElement([RefundStatus::STATUS_QUEUED, RefundStatus::STATUS_PENDING, RefundStatus::STATUS_FAILED, RefundStatus::STATUS_PROCESSING, RefundStatus::STATUS_REFUNDED]);
```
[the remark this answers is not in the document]
schmitzcarsten: forgot this?

schmitzcarsten (changes requested):

.phpstan.lvl8.neon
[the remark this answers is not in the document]
KienerNL: Yes, is removed now

tests/PHPUnit/Hydrator/RefundHydratorTest.php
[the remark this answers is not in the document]
KienerNL: Every status does not need a test case. The test is to see if the hydrator correctly returns a Refund object as an array. What test data we use for the Refund object and the expected array is largely irrelevant, but the data in both the expected array as well as the Refund object has to be the same in each separate test case. That's why both the array en Refund object are created in the same function.

src/Service/MollieApi/Order.php
KienerNL: Authorized is correct here. This is a `Mollie\Api\Resources\Payment` which can only return statusses from `Mollie\Api\Types\PaymentStatus`

boxblinkracer (approved):

schmitzcarsten (approved):

schmitzcarsten (approved):

PR #146, opened 2021-06-11
MOL-453: Refactor Webhook Notifications

boxblinkracer:
completely refactored the webhooks in order to correctly update all payment status possibilities

### DIFF
--- a/src/Facade/Notifications/NotificationFacade.php
+++ b/src/Facade/Notifications/NotificationFacade.php
@@ -1,0 +1,198 @@
+<?php
+
+namespace Kiener\MolliePayments\Facade\Notifications;
+
+use Kiener\MolliePayments\Gateway\MollieGatewayInterface;
+use Kiener\MolliePayments\Handler\Method\ApplePayPayment;
+use Kiener\MolliePayments\Service\Mollie\OrderStatusConverter;
+use Kiener\MolliePayments\Service\Order\OrderStatusUpdater;
+use Kiener\MolliePayments\Setting\MollieSettingStruct;
+use Mollie\Api\Resources\Order;
+use Shopware\Core\Checkout\Cart\Exception\OrderNotFoundException;
+use Shopware\Core\Checkout\Order\Aggregate\OrderTransaction\OrderTransactionEntity;
+use Shopware\Core\Checkout\Order\OrderEntity;
+use Shopware\Core\Checkout\Payment\PaymentMethodEntity;
+use Shopware\Core\Framework\Context;
+use Shopware\Core\Framework\DataAbstractionLayer\EntityRepositoryInterface;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\ContainsFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\MultiFilter;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+
+
+class NotificationFacade
+{
+
+    /**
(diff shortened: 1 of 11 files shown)


boxblinkracer:
requires personal review with @schmitzcarsten first

barbieswimcrew (approved):

src/Facade/Notifications/NotificationFacade.php
```diff
@@ -0,0 +1,198 @@
...
+     * @return OrderTransactionEntity
+     */
+    private function getTransaction(string $transactionId, Context $context): OrderTransactionEntity
+    {
+        $criteria = new Criteria();
+        $criteria->addFilter(new EqualsFilter('id', $transactionId));
+        $criteria->addAssociation('order');
+        $criteria->addAssociation('paymentMethod');
+
+        return $this->repoOrderTransactions->search($criteria, $context)->first();
```
schmitzcarsten, line 138: value could be null, but method should only return OrderTransactionEntity. Make it nullable

src/Gateway/Mollie/MollieGatewayFactory.php
```diff
@@ -0,0 +1,22 @@
+<?php
```
schmitzcarsten, line 1: are you using this factory anywhere? If not, please delete. I think it is over engineered, isn't it?

src/Resources/config/services/gateways.xml
```diff
@@ -0,0 +1,17 @@
+<?xml version="1.0" ?>
+
+<container xmlns="http://symfony.com/schema/dic/services"
+           xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+           xsi:schemaLocation="http://symfony.com/schema/dic/services http://symfony.com/schema/dic/services/services-1.0.xsd">
+
+    <services>
+
+        <service id="Kiener\MolliePayments\Gateway\Mollie\MollieGatewayFactory">
```
schmitzcarsten, line 9: think it could be removed here too

src/Service/Order/OrderStatusUpdater.php
```diff
@@ -0,0 +1,201 @@
...
+                    if (defined('Shopware\Core\System\StateMachine\Aggregation\StateMachineTransition\StateMachineTransitionActions::ACTION_AUTHORIZE')) {
+                        $transitionTarget = PaymentStatus::STATUS_AUTHORIZED;
+                        $transitionAction = StateMachineTransitionActions::ACTION_AUTHORIZE;
+                    } else {
+                        $transitionTarget = PaymentStatus::STATUS_PAID;
+                        $transitionAction = StateMachineTransitionActions::ACTION_PAY;
+                    }
+
+                    if ($currentStatus !== $transitionTarget) {
+                        $transition = new Transition(OrderTransactionDefinition::ENTITY_NAME, $transaction->getId(), $transitionAction, 'stateId');
```
schmitzcarsten, line 98: I think, we should add an authorize action in the transition handler and use it here

What about a transition that would not be possible/forbidden. We could add this logic to the transitionhandler too

schmitzcarsten (changes requested):

src/Gateway/Mollie/MollieGatewayFactory.php
boxblinkracer: yes perfect, is now not used anymore since our latest changes, thx

src/Service/Order/OrderStatusUpdater.php
boxblinkracer: this transition handler is the shopware transition handler
i would now leave it this way and then it it to the handler that you are building

schmitzcarsten: i see, that's why I want mine merged. Brain fuck up where the classes exist...

schmitzcarsten (approved):

PR #149, opened 2021-06-17
MOL-461: Fix iDEAL Issuers in Shopware 6.4

boxblinkracer:
added compatible code for shopware >= 6.4 and <

the new code will change the selected iDeal issuer on dropdown value changed

the old code somehow didnt work in 6.3.5.2, so i've changed the event handler from a "form listener" to a "button click" listener, which worked

i also added 2 cypress tests that evaluate that a pre-selected issuer is being sent with the payment request.
in that case, there will immediately be a payment status form instead of an additional issuer-selection list in Mollie
that asserts that an existing issuer has been set


boxblinkracer:
please note this is based on https://github.com/mollie/Shopware6/pull/132
it finalizes and fixes this other PR

schmitzcarsten (approved):

PR #151, opened 2021-06-21
MOL-353: refactor payment handler

schmitzcarsten:


src/Exception/MissingPriceLineItem.php
```diff
@@ -7,7 +7,7 @@
 use Shopware\Core\Framework\ShopwareHttpException;
 use Symfony\Component\HttpFoundation\Response;
 
-class MissingPriceLineItemException extends ShopwareHttpException
+class MissingPriceLineItem extends ShopwareHttpException
```
boxblinkracer, line 10: shouldnt it be named exception? accidentally renamed?

src/Exception/MollieOrderPaymentCouldNotBeCreated.php
```diff
@@ -0,0 +1,30 @@
...
+class MollieOrderPaymentCouldNotBeCreated extends ShopwareHttpException
+{
+    /**
+     * @param string $id
+     * @param array<string,mixed> $parameters
+     * @param \Throwable|null $previous
+     */
+    public function __construct(string $id, array $parameters = [], \Throwable $previous = null)
+    {
+        $message = sprintf('Could not create a new payment for mollie order(%s)', $id);
```
boxblinkracer, line 17: i think a space is missing between order and id...just a small hint

src/Exception/PaymentUrlException.php
```diff
@@ -8,12 +8,18 @@
...
 {
 
-    /**
-     * @return string
-     */
+    public function __construct(string $orderTransactionId, string $errorMessage)
+    {
+        parent::__construct(
+            $orderTransactionId,
+            'Could not create a Mollie payment url due to the following error:' . \PHP_EOL . '{{ errorMessage }}',
```
boxblinkracer, line 15: facing the topic "logging", im not quite sure if a forced new line is good?

src/Facade/MolliePaymentDoPay.php
```diff
@@ -0,0 +1,181 @@
...
+     */
+    public function preparePayProcessAtMollie(
+        string $paymentMethod,
+        AsyncPaymentTransactionStruct $transactionStruct,
+        SalesChannelContext $salesChannelContext,
+        PaymentHandler $paymentHandler
+    ): string
+    {
+        // get order with all needed associations
+        $order = $this->orderService->getOrder($transactionStruct->getOrder()->getId(), $salesChannelContext->getContext());
```
boxblinkracer, line 113: i guess the order in the transaction is already an entity object? wouldnt that one be already NULL if its not existing and lead to an exception when calling ->getId()?

src/Facade/MolliePaymentDoPay.php
```diff
@@ -0,0 +1,181 @@
...
+            $this->logger->addDebugEntry(
+                'Found an existing mollie order with id %s.',
+                $salesChannelContext->getSalesChannelId(),
+                $salesChannelContext->getContext()
+            );
+
+            $payment = $this->apiOrderService->createOrReusePayment($mollieOrderId, $paymentMethod, $salesChannelContext);
+
+            // if direct payment return to success page
+            if ($payment->isPaid()) {
```
boxblinkracer, line 135: why returning if its paid?
im missing 
a) a check for an empty checkout url
b) additional successful payment status (MolliePaymentStatus::isValidPayment

i would expect a check if its a valid payment , but still has no checkout URL then the return URL is called immediately

src/Facade/MolliePaymentDoPay.php
```diff
@@ -0,0 +1,181 @@
...
+
+        if ($mollieOrder instanceof MollieOrder) {
+            $customFieldsStruct->setMollieOrderId($mollieOrder->id);
+            $customFieldsStruct->setMolliePaymentUrl($mollieOrder->getCheckoutUrl());
+
+            $this->updateOrderCustomFields->updateOrder($order->getId(), $customFieldsStruct, $salesChannelContext);
+            $this->updateOrderLineItems->updateOrderLineItems($mollieOrder, $salesChannelContext);
+        }
+
+        return $customFieldsStruct->getMolliePaymentUrl() ?? $customFieldsStruct->getTransactionReturnUrl() ?? $transactionStruct->getReturnUrl();
```
boxblinkracer, line 179: i think that code seems legit
still...do you think its possible to have 1 source of truth with all that payment and checkout URL stuff.
for me it seems as if the order creation is optional and depending on if its existing already.
so the order creation could be moved up, and then it all uses the same code for URL handling?

src/Facade/MolliePaymentFinalize.php
```diff
@@ -0,0 +1,89 @@
...
+
+        $paymentStatus = $this->paymentStatusHelper->processPaymentStatus(
+            $transactionStruct->getOrderTransaction(),
+            $order,
+            $mollieOrder,
+            $salesChannelContext->getContext()
+        );
+
+        if ($paymentStatus === PaymentStatus::STATUS_CANCELED || $paymentStatus === PaymentStatus::STATUS_FAILED) {
+            $this->transactionTransitionService->reOpenTransaction(
```
boxblinkracer, line 74: wouldnt that make more sense before processing it?
also , there would be a function MolliePaymentStatus::isFailedPayment
that handles cancelled, failed and expired...so its basically an indicator if a payment can be continued or not...maybe it can be used in here?

src/Handler/Method/ApplePayPayment.php
```diff
@@ -5,30 +5,28 @@
...
         SalesChannelContext $salesChannelContext,
-        CustomerEntity $customer,
-        LocaleEntity $locale
+        CustomerEntity $customer
     ): array
     {
-        if (!array_key_exists(static::FIELD_APPLE_PAY_PAYMENT_TOKEN, $orderData[static::FIELD_PAYMENT]) || in_array($orderData[static::FIELD_PAYMENT][static::FIELD_APPLE_PAY_PAYMENT_TOKEN], [null, ''], true)) {
+        $applePayToken = $orderData['payment']['applePayPaymentToken'] ?? '';
+
+        if (empty($applePayToken)) {
```
boxblinkracer, line 26: shouldnt there be the api token used for apple pay direct?

src/Handler/Method/CreditCardPayment.php
```diff
@@ -16,21 +20,43 @@ class CreditCardPayment extends PaymentHandler
...
+
+        if (empty($cardToken)) {
+            return $orderData;
+        }
+
+        $orderCardToken = $orderData['payment']['cardToken'] ?? '';
 
-        if (!array_key_exists(static::FIELD_CREDIT_CARD_TOKEN, $orderData[static::FIELD_PAYMENT]) || in_array($orderData[static::FIELD_PAYMENT][static::FIELD_CREDIT_CARD_TOKEN], [null, ''], true)) {
-            $orderData[static::FIELD_PAYMENT][static::FIELD_CREDIT_CARD_TOKEN] = $customFields['mollie_payments']['credit_card_token'];
+        if (empty($orderCardToken)) {
```
boxblinkracer, line 55: if ! empty? + wrong variable used? $orderCardToken vs $cartToken?

src/Handler/Method/DirectDebitPayment.php
```diff
@@ -5,30 +5,28 @@
...
         array $orderData,
         SalesChannelContext $salesChannelContext,
-        CustomerEntity $customer,
-        LocaleEntity $locale
+        CustomerEntity $customer
     ): array
     {
-        if (!array_key_exists(static::FIELD_SEPA_DIRECT_DEBIT_CONSUMER_NAME, $orderData[static::FIELD_PAYMENT]) || in_array($orderData[static::FIELD_PAYMENT][static::FIELD_SEPA_DIRECT_DEBIT_CONSUMER_NAME], [null, ''], true)) {
-            $orderData[static::FIELD_PAYMENT][self::FIELD_SEPA_DIRECT_DEBIT_CONSUMER_NAME] = sprintf('%s %s', $customer->getFirstName(), $customer->getLastName());
+        $consumerName = $orderData['payment']['consumerName'] ?? '';
```
boxblinkracer, line 24: how can it be that a consumer name as "payment specific data" is already set in here?
i would have thought of setting it, either from source A or B
but its overwritten if its empty...but why is it even filled before?

src/Handler/Method/iDealPayment.php
```diff
@@ -17,15 +16,23 @@ class iDealPayment extends PaymentHandler
...
+
+        $issuer = $customFields['mollie_payments']['preferred_ideal_issuer'] ?? '';
+        $paymentIssuer = $orderData['payment']['issuer'] ?? '';
+
+        if (empty($issuer)) {
+            return $orderData;
+        }
+
+        if (empty($paymentIssuer)) {
+            $orderData['payment']['issuer'] = $issuer;
```
boxblinkracer, line 35: why does it have 2 preselected issuers? field A is either filed or not? is there an additional feature existing?

src/Handler/PaymentHandler.php
```diff
@@ -161,144 +89,64 @@ protected function processPaymentMethodSpecificParameters(array $orderData, Sale
...
-         */
         try {
-            $this->setApiKeysBySalesChannelContext($salesChannelContext);
-        } catch (Exception $e) {
+            $paymentUrl = $this->payFacade->preparePayProcessAtMollie($this->paymentMethod, $transaction, $salesChannelContext, $this);
+        } catch (Throwable $exception) {
+            $logException = null;
+            $logLevel = Logger::CRITICAL;
+
+            if ($exception instanceof \Exception) {
```
boxblinkracer, line 111: you could do a catch Exception and a catch Throwable in this case...maybe it helps to distinguish

tests/PHPUnit/Service/MollieApi/Builder/MollieOrderBuilderTest.php
```diff
@@ -0,0 +1,1005 @@
...
+            'shippingAddress' => $this->getExpectedTestAddress($this->address, $this->email),
+            'expiresAt' => $expectedOrderLifeTime
+        ];
+
+        self::assertSame($expected, $actual);
+    }
+
+    public function testBanContactOrder(): void
+    {
+        $redirectWebhookUrl = 'https://foo';
```
boxblinkracer, line 220: would it make sense to extract payment methods to separate files?

tests/PHPUnit/Setting/MollieSettingsStructTest.php
```diff
@@ -107,8 +106,8 @@ public function orderBankTransferDueDays()
         $today = (new DateTime())->setTimezone(new DateTimeZone('UTC'));
 
         return [
-            'dueDateDays cannot be smaller than minimum' => [-1, BankTransferPayment::DUE_DATE_MIN_DAYS, (clone $today)->modify('+1 day')->format('Y-m-d')],
-            'dueDateDays cannot be bigger than maximum' => [1000, BankTransferPayment::DUE_DATE_MAX_DAYS, (clone $today)->modify('+100 day')->format('Y-m-d')],
+            //'dueDateDays cannot be smaller than minimum' => [-1, BankTransferPayment::DUE_DATE_MIN_DAYS, (clone $today)->modify('+1 day')->format('Y-m-d')],
```
boxblinkracer, line 109: accidentally removed?

src/Exception/MollieOrderPaymentCouldNotBeCreated.php
schmitzcarsten: added

src/Exception/PaymentUrlException.php
schmitzcarsten: removed

src/Facade/MolliePaymentDoPay.php
schmitzcarsten: it's an AsyncPaymentTransactionStruct. It can only be constructed with an order entity. The order service makes an db lookup and gets an order entity with more associations. The return value of the service could be null. So yes it's nearly impossible to get null from service, but it is not impossible. And stan would complain without check ;)

src/Facade/MolliePaymentDoPay.php
schmitzcarsten: no first there is a check if a payment is already done. There we check if an open payment exists and can be reused or it is cancelled and created new. Url handling is a different thing here and we will return in this case the correct url.

And at the bottom there is no payment at all. We are creating a mollie order and take the url from there or use shopware standard url, if a direct payment has been used

tests/PHPUnit/Setting/MollieSettingsStructTest.php
schmitzcarsten: yes readded

src/Handler/Method/ApplePayPayment.php
schmitzcarsten: token is added in controller and used here

src/Handler/Method/CreditCardPayment.php
schmitzcarsten: refactored

src/Handler/Method/DirectDebitPayment.php
schmitzcarsten: refactored

src/Handler/Method/iDealPayment.php
schmitzcarsten: as discussed togeather, left unchanged and will be refactored in future

tests/PHPUnit/Service/MollieApi/Builder/MollieOrderBuilderTest.php
schmitzcarsten: done

PR #158, opened 2021-07-01
MOL-504: add backwards compatibility

schmitzcarsten:
in SW versions < 6.3 there is no method getSalesChannelId in SalesChannelContext

### DIFF
--- a/src/Facade/SetMollieOrderRefunded.php
+++ b/src/Facade/SetMollieOrderRefunded.php
@@ -3,6 +3,7 @@
 namespace Kiener\MolliePayments\Facade;
 
 use Kiener\MolliePayments\Exception\CouldNotSetRefundAtMollieException;
+use Kiener\MolliePayments\Exception\MissingSalesChannelInOrder;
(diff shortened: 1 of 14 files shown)


src/Facade/SetMollieOrderRefunded.php
```diff
@@ -70,7 +72,13 @@ public function setRefunded(string $orderTransactionId, Context $context): void
             );
         }
 
-        $apiClient = $this->apiFactory->getClient($order->getSalesChannelId(), $context);
+        $salesChannel = $order->getSalesChannel();
+
+        if (!$salesChannel instanceof SalesChannelEntity) {
+            throw new MissingSalesChannelInOrder($order->getOrderNumber() ?? $order->getId());
```
boxblinkracer, line 78: argument should be order number

so here you have either order number or id? is that alright?

schmitzcarsten: yes, theoretical orderNumber could be null. So I added a check that it gets order id in these theoretic cases

barbieswimcrew (approved):

PR #169, opened 2021-07-23
[MOL-486] Create customers at Mollie

KienerNL:


src/Service/MollieApi/MollieOrderCustomerEnricher.php
```diff
@@ -9,7 +9,7 @@ class MollieOrderCustomerEnricher
 {
     public function enrich(array $orderData, CustomerEntity $customer, MollieSettingStruct $settings): array
     {
-        if ($settings->createNoCustomersAtMollie()) {
+        if (!$settings->createCustomersAtMollie()) {
             return $orderData;
         }
```
schmitzcarsten, line 15: pls add a todo here we have to adjust the following code.

And as we have no implementation yet, always return $orderData

src/Service/SettingsService.php
```diff
@@ -43,4 +43,25 @@ public function getSettings(?string $salesChannelId = null, ?Context $context =
...
+    {
+        $this->systemConfigService->set(self::SYSTEM_CONFIG_DOMAIN . $key, $value, $salesChannelId);
+    }
+
+    public function delete(string $key, ?string $salesChannelId = null): void
+    {
+        $this->systemConfigService->delete(self::SYSTEM_CONFIG_DOMAIN . $key, $salesChannelId);
+    }
+
+    public function setProfileId(?string $profileId, ?string $salesChannelId = null, bool $testMode = false): void
```
schmitzcarsten, line 57: may saleschannel be null in any case?

src/Subscriber/SystemConfigSubscriber.php
```diff
@@ -0,0 +1,123 @@
...
+    public function __construct(SettingsService $settingsService)
+    {
+        $this->settingsService = $settingsService;
+        $this->apiClient = new MollieApiClient();
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            'system_config.written' => 'onSystemConfigWritten',
```
schmitzcarsten, line 32: what other event possibilities are here ? Using dal events is really the last way

src/Subscriber/SystemConfigSubscriber.php
```diff
@@ -0,0 +1,123 @@
...
+                $payload['configurationKey'],
+                $payload['configurationValue'],
+                $payload['salesChannelId']
+            );
+        }
+    }
+
+    private function checkSystemConfigChange(string $key, $value, ?string $salesChannelId = null)
+    {
+        if (in_array($key, [
```
schmitzcarsten, line 50: if dal event has to be used: We only have to do a call if the api keys change. Here you check if the profile id changes. This does not feel correct

src/Service/CustomerService.php
```diff
@@ -222,24 +238,68 @@ public function setIDealIssuer(CustomerEntity $customer, string $issuerId, Conte
...
+    }
+
+    /**
+     * @param string $customerId
+     * @param string $mollieCustomerId
+     * @param string $profileId
+     * @param bool $testMode
+     * @param Context $context
+     */
+    public function setMollieCustomerId(
```
schmitzcarsten, line 262: I'm missing unit tests for the new struct and the interaction with the legacy customerID.

But this is one of the most important changes of this ticket. Please add them.

src/Subscriber/SystemConfigSubscriber.php
```diff
@@ -0,0 +1,123 @@
...
+
+    /** @var MollieApiClient */
+    private $apiClient;
+
+    /** @var array */
+    private $profileIdStorage = [];
+
+    public function __construct(SettingsService $settingsService)
+    {
+        $this->settingsService = $settingsService;
```
schmitzcarsten, line 25: add tests for subscriber

schmitzcarsten (changes requested):

src/Service/CustomerService.php
```diff
@@ -222,24 +238,68 @@ public function setIDealIssuer(CustomerEntity $customer, string $issuerId, Conte
...
      */
-    public function getCustomer(string $customerId, Context $context) : ?CustomerEntity
+    public function getCustomer(string $customerId, Context $context): ?CustomerEntity
     {
         $customer = null;
 
         try {
             $criteria = new Criteria();
             $criteria->addFilter(new EqualsFilter('id', $customerId));
-            $criteria->addAssociation('activeShippingAddress');
```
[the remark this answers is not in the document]
KienerNL: These associations are already being added in the `$criteria->addAssociations` call below. I deleted these because they are duplicate.

src/Subscriber/SystemConfigSubscriber.php
KienerNL: In 6.4 they added an `SystemConfigChangedEvent` that we could use, but then the checks for 6.3.5.4 would fail. The only way I can see is using the dal event until support for 6.3 is dropped.

src/Subscriber/SystemConfigSubscriber.php
KienerNL: This is really tricky. On first setup a profileId is not set yet in the system config. Then everything works fine and as expected.
However once a profileId is set in the system config, even if there is no field in the plugin's configuration that uses the profileId field, the Shopware plugin config page will still send that value to be saved to the database again.
image

Because now it saves an old profileId, I have to check whether we have already gotten a new profileId after saving the api key, and then save that new key again.

src/Service/SettingsService.php
KienerNL: Yes, when salesChannelId is null, the profileId will be saved for the default global settings

src/Subscriber/OrderPlacedSubscriber.php
```diff
@@ -0,0 +1,85 @@
...
+        $this->customerService = $customerService;
+        $this->customerApiService = $customerApiService;
+        $this->extractor = $extractor;
+        $this->settingsService = $settingsService;
+        $this->logger = $logger;
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [
```
schmitzcarsten, line 50: We can't do it here in a subscriber.

If an order is placed, the asynchronous payment process begins. Here we create the mollieOrder Struct and in this process we have to create the customer first before placing the order at mollie. 

Theoretical it is possible that the process creating mollie order struct is faster than the event listening here. In these cases we couldn't add the mollie customer id to the mollie order struct.

The correct place/time for creating the mollie customer is in MolliePaymentDoPay Facade. Just before creating the mollie order struct in line 159 ($mollieOrderArray = $this->orderBuilder->build)

If configuration says we should create an user, create it. The first line in the builder build method fetches the customer from database. So we have the information there and only need to query the custom fields and the settings of course

schmitzcarsten (changes requested):

src/Subscriber/OrderPlacedSubscriber.php
KienerNL: But the event we're currently listening for is being triggered from `CheckoutController::order` on line 188, in the createOrder process from the orderService, and the asynchronous payment process isn't triggered until line 203 from that same function. So it's always going to hit the subscriber before starting payment

KienerNL:
PHPStan checks should fail now on `src/Service/Transition/TransactionTransitionService.php:167` because `STATE_AUTHORIZED` did not exist in 6.3.5.4. Related commit e34f21e10c180d1da87bfa7eecaebc52989935a7

PR #200, opened 2021-09-17
Use RouterInterface instead of concrete Router class

alexnbio:
-Use RouterInterface instead of concrete Router class

### DIFF
--- a/src/Service/ShopService.php
+++ b/src/Service/ShopService.php
@@ -2,15 +2,15 @@
 
 namespace Kiener\MolliePayments\Service;
 
-use Shopware\Storefront\Framework\Routing\Router;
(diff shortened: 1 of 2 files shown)


boxblinkracer:
Hi
thank you very much for this

just a small question, do you have any real use case for this, or is it just an improvement to make it better (yes i AM a fan of interfaces :) )

alexnbio:
Hey, 
yes, we have an issue because of another plugin that decorated the Router class from Shopware and now Mollie throws 500 errors in those two classes I submitted the change for.

Thank you for responding so quick :).

boxblinkracer:
hi, perfect thank you
merged and added to upcoming release plan :)

thx a lot for contributing

PR #202, opened 2021-10-07
NTR: Remove duplicate tests

KienerNL:
While writing tests for MOL-200 I noticed there were two files containing the same tests. One was `/Service/**Order**/OrderTransitionServiceTest.php`, created by Thilo on April 16th, and the second was `/Service/**Transition**/OrderTransitionServiceTest.php`, created by Carsten on July 12th during the payment handler refactor.

Thilo's test file contained one more additional test, which I added to Carsten's file. The only other differences at that point we're two extra unused services in the setUp function, in Thilo's test file. The rest are duplicated fragments.

image


boxblinkracer:
hey
great work, you are absolutely correct
thanks for this :)

PR #203, opened 2021-10-08
MOL-200: Shipment HTTP API

KienerNL:
Two new api endpoints have been added for shipping orders.

`/api/mollie/ship/order`
`[GET] number` ordernumber of the order to ship
Allows to ship an order using the order number. Any unshipped line items will be shipped.

`/api/mollie/ship/item`
`[GET] order` ordernumber of the order to ship
`[GET] item` identifier for the line item to ship ¹
`[GET] quantity (optional)` The amount of the line item to ship, if none is passed the entire line item is shipped.

¹ Shopware 6 does not have a human readable number for each line item like Shopware 5 has. Therefore, several options are available for this endpoint.
- Use the product number if the line item is of type "product".
- The Mollie line id.
- The internal Shopware id of the entity the line item is referencing.
- The internal Shopware id of the line item itself.

### DIFF
--- a/tests/PHPUnit/Service/OrderServiceTest.php
+++ b/tests/PHPUnit/Service/OrderServiceTest.php
@@ -18,6 +18,8 @@ use Shopware\Core\Checkout\Order\SalesChannel\OrderService as ShopwareOrderServi
 use Shopware\Core\Framework\Context;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\Criteria;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\EntitySearchResult;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\EqualsFilter;
+use Shopware\Core\Framework\DataAbstractionLayer\Search\Filter\Filter;
 use Shopware\Core\Framework\DataAbstractionLayer\Search\IdSearchResult;
 
 class OrderServiceTest extends TestCase
@@ -106,6 +108,16 @@ class OrderServiceTest extends TestCase
         $receivedCriteria = $this->orderRepository->criteria[0];
 
         $this->assertTrue($receivedCriteria->hasEqualsFilter('orderNumber'));
+
+        $orderNumberFilters = array_filter($receivedCriteria->getFilters(), static function (Filter $filter) {
+            return $filter instanceof EqualsFilter && $filter->getField() === 'orderNumber';
+        });
+
+        $orderNumberFilterValues = array_map(static function (EqualsFilter $filter) {
+            return $filter->getValue();
+        }, $orderNumberFilters);
+
+        $this->assertContains('bar', $orderNumberFilterValues);
     }
 
     public function testGetOrderByNumberDoesntExist()


src/Facade/MollieShipment.php
```diff
@@ -120,4 +159,112 @@ public function setShipment(string $orderDeliveryId, Context $context): bool
...
+    public function shipItem(string $orderNumber, string $itemIdentifier, int $quantity, Context $context): \Mollie\Api\Resources\Shipment
+    {
+        $order = $this->orderService->getOrderByNumber($orderNumber, $context);
+
+        $mollieOrderId = $this->orderService->getMollieOrderId($order);
+
+        $lineItems = $this->searchLineItem($order, $itemIdentifier, $context);
+
+        if ($lineItems->count() > 1) {
+            throw new OrderLineItemFoundManyException($itemIdentifier);
```
boxblinkracer, line 187: great idea :)

src/Facade/MollieShipment.php
```diff
@@ -120,4 +159,112 @@ public function setShipment(string $orderDeliveryId, Context $context): bool
...
+        if ($this->mollieApiOrderService->isCompletelyShipped($mollieOrderId, $order->getSalesChannelId(), $context)) {
+            $this->deliveryTransitionService->shipDelivery($delivery, $context);
+        } else {
+            $this->deliveryTransitionService->partialShipDelivery($delivery, $context);
+        }
+
+        return $shipment;
+    }
+
+    public function searchLineItem(OrderEntity $order, string $itemIdentifier, Context $context): OrderLineItemCollection
```
boxblinkracer, line 227: i guess this should be private?
i like the idea of this function, still struggling a bit if it might be a better decision to have functions like getByProductNumber, getByMollieID.
i'm a bit afraid of it grow over the time and then its a function that just does "anything" from an outside scope.
what do you think?
but at least for now i'm fine with it

src/Service/MollieApi/Order.php
```diff
@@ -211,12 +231,38 @@ public function setShipment(string $mollieOrderId, string $salesChannelId, Conte
...
+     * @throws CouldNotFetchMollieOrderException
+     */
+    public function isCompletelyShipped(string $mollieOrderId, string $salesChannelId, Context $context): bool
+    {
+        $mollieOrder = $this->getMollieOrder($mollieOrderId, $salesChannelId, $context);
+
+        /** @var OrderLine $mollieOrderLine */
+        foreach ($mollieOrder->lines() as $mollieOrderLine) {
+            if ($mollieOrderLine->shippableQuantity > 0 &&
+                in_array($mollieOrderLine->type, [
```
boxblinkracer, line 248: question regarding this in_array
do you think that could create problems with 3rd party plugins?
just to double check.
promotions would always be marked as shipped right? which is a nice thing, because thy cannot be shipped

src/Exception/OrderDeliveryNotFound.php
```diff
@@ -0,0 +1,30 @@
+<?php declare(strict_types=1);
+
+namespace Kiener\MolliePayments\Exception;
+
+use Shopware\Core\Framework\ShopwareHttpException;
+use Symfony\Component\HttpFoundation\Response;
+
+class OrderDeliveryNotFound extends ShopwareHttpException
```
boxblinkracer, line 8: could you add Exception suffixes to all exceptions.
some do have them, some not...just to be consisent, thanks

src/Service/MollieApi/Order.php
KienerNL: Do you mean 3rd party plugins that rely on or reuse the Mollie plugin? Or just 3rd party plugins in general?

src/Facade/MollieShipment.php
KienerNL: Hmmm, could be done. But suppose `getByProductNumber` and `getByMollieId` both return one lineItem. (Sidenote: that should never happen, but let's pretend it does.) We'll then have two collections that first have to be merged.

Are there any other unique identifiers on a line item that could cause this function to grow? The only ones I can see right now is indeed the product number, and the internal ids of the entities, as well as the mollie order line id. All other current data is either not unique, or not human readable.

Perhaps renaming the function would make it clearer what it should be used for.

src/Struct/MollieApi/ShipmentTrackingInfoStruct.php
```diff
@@ -0,0 +1,87 @@
+<?php
+
+namespace Kiener\MolliePayments\Struct\MollieApi;
+
+use Shopware\Core\Framework\Struct\Struct;
+
+class ShipmentTrackingInfoStruct extends Struct
```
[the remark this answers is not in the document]
KienerNL: A relic from day 1 ;)

I first had added a way to update the tracking info as well, but decided to leave that out to keep some consistency with the shopware 5 plugin. Evidently I forgot a file.

src/Exception/OrderDeliveryNotFound.php
boxblinkracer: i mean...go ahead :D

src/Service/MollieApi/Order.php
boxblinkracer: plugins that create new types of products? those would be marked as "shipped" which seems fine for me

src/Exception/OrderDeliveryNotFound.php
KienerNL: Will create a new PR shortly

PR #206, opened 2021-10-15
NTR: Rename exceptions to add Exception suffix

KienerNL:
As requested in https://github.com/mollie/Shopware6/pull/203#discussion_r729037084


src/Exception/CouldNotFetchTransactionException.php
```diff
@@ -5,7 +5,7 @@
 use Shopware\Core\Framework\ShopwareHttpException;
 use Symfony\Component\HttpFoundation\Response;
 
-class CouldNotFetchTransaction extends ShopwareHttpException
+class CouldNotFetchTransactionException extends ShopwareHttpException
```
schmitzcarsten, line 8: Why do you rename with suffix? The namespace has Exception in its path

PR #208, opened 2021-10-19
MOL-216: Show Vouchers in Refund form in Admin

boxblinkracer:
show voucher value in refund view, so that the refundable amount "makes sense"
because vouchers cannot be refunded

### DIFF
--- a/src/Service/RefundService.php
+++ b/src/Service/RefundService.php
@@ -36,8 +36,8 @@ class RefundService
      * @param RefundHydrator $refundHydrator
      */
     public function __construct(
-        Order $mollieOrderApi,
-        OrderService $orderService,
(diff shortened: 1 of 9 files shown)


src/Service/RefundService.php
```diff
@@ -164,6 +164,40 @@ public function getRemainingAmount(OrderEntity $order, Context $context): float
...
+
+        if (!property_exists($payment->details, 'vouchers')) {
+            return 0;
+        }
+
+        $voucherAmount = 0;
+
+        /** @var \stdClass $voucher */
+        foreach ($payment->details->vouchers as $voucher) {
+            # weird but we need that cast and separate variable
```
uehler, line 192: because of phpstan?

PR #222, opened 2021-11-02
Add check for refunds array if it is not undefined

zaifastafa:
In a shop for a client, I was having an issue when using mollie credit card payment that the order items in the order details was not showing up in the admin section. The error was giving an "undefined" error at this point when reading the length property. Adding an additional check here for refunds array if it is available before checking for length solves the problem.

image


boxblinkracer:
Hi

thank you so much for this fix.
definitley going to add it

please give us a bit, we'll create a ticket and so on 

thank you so much for helping to improve the quality of the plugin

PR #226, opened 2021-11-03
MOL-635: Failed payments have status cancelled instead of failed

boxblinkracer:
with the shopware default behaviour for failed payments,
the failed payments lead to a status "cancelled".

so we only do this if its really cancelled,
otherwise it should be an exception that leads to the status "failed"

### DIFF
--- a/src/Facade/MolliePaymentFinalize.php
+++ b/src/Facade/MolliePaymentFinalize.php
@@ -13,6 +13,7 @@ use Kiener\MolliePayments\Struct\MollieOrderCustomFieldsStruct;
 use Mollie\Api\Exceptions\ApiException;
 use Mollie\Api\Exceptions\IncompatiblePlatform;
 use Shopware\Core\Checkout\Payment\Cart\AsyncPaymentTransactionStruct;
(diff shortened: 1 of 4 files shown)


src/Facade/MolliePaymentFinalize.php
```diff
@@ -76,21 +77,49 @@ public function finalize(AsyncPaymentTransactionStruct $transactionStruct, Sales
...
+        # now either set the payment status for successful payments
+        # or make sure to throw an exception for Shopware in case
+        # of failed payments.
+        if (!MolliePaymentStatus::isFailedStatus($paymentStatus)) {
+
+            $this->orderStatusUpdater->updatePaymentStatus($transactionStruct->getOrderTransaction(), $paymentStatus, $salesChannelContext->getContext());
+
+        } else {
+
+            $orderTransactioID = $transactionStruct->getOrderTransaction()->getUniqueIdentifier();
```
schmitzcarsten, line 107: typo missing n at $orderTransactio  n   ID

schmitzcarsten (changes requested):

PR #228, opened 2021-11-04
NTR: Ignore compiled admin css

KienerNL:
Similar to the compiled JS, the compiled CSS should also not be in the repository, but should be built before preparing a release.


boxblinkracer:
perfect thanks

PR #232, opened 2021-11-09
MOL-217: Add Klarna Pay Now

boxblinkracer:


barbieswimcrew (approved):

composer.lock
```diff
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
```
schmitzcarsten, line 4: is this only a composer update? No composer.json with changes

schmitzcarsten (approved):

composer.lock
boxblinkracer: you're right
i forgot that file thanks
requires a new api framework for the constant :)

PR #240, opened 2021-11-26
MOL-623: Change selector for confirm button

KienerNL:
The account order edit page does not have an id attribute on it's submit button, so the components would not get initialized.

### DIFF
--- a/src/Resources/app/storefront/src/mollie-payments/plugins/creditcard-components-sw64.plugin.js
+++ b/src/Resources/app/storefront/src/mollie-payments/plugins/creditcard-components-sw64.plugin.js
@@ -88,7 +88,7 @@ export default class MollieCreditCardComponentsSw64 extends Plugin {
             mollieController: 'div.mollie-components-controller',
             paymentForm: '#changePaymentForm',
             confirmForm: '#confirmOrderForm',
-            confirmFormButton: '#confirmFormSubmit',
+            confirmFormButton: '#confirmOrderForm > button[type="submit"]',
         };
     }
 
--- a/tests/Cypress/cypress/integration/storefront/payment-methods/creditcard.spec.js
+++ b/tests/Cypress/cypress/integration/storefront/payment-methods/creditcard.spec.js
@@ -158,7 +158,7 @@ describe('Credit Card Components', () => {
                 assertComponentErrors(true, true, true, false);
             })
 
-            it('Components are initialized after failed payment', () => {
+            it('Complete payment after failure with components', () => {
                 // We need to test this with Shopware's complete order page.
                 configAction.setupShop(false, true, false);
 
--- a/tests/Cypress/cypress/integration/storefront/payment-methods/creditcard.spec.js
+++ b/tests/Cypress/cypress/integration/storefront/payment-methods/creditcard.spec.js
@@ -177,12 +177,7 @@ describe('Credit Card Components', () => {
 
                 cy.url().should('include', 'https://www.mollie.com/checkout/');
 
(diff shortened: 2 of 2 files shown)


src/Resources/app/storefront/src/mollie-payments/plugins/creditcard-components-sw64.plugin.js
```diff
@@ -88,7 +88,7 @@ export default class MollieCreditCardComponentsSw64 extends Plugin {
             mollieController: 'div.mollie-components-controller',
             paymentForm: '#changePaymentForm',
             confirmForm: '#confirmOrderForm',
-            confirmFormButton: '#confirmFormSubmit',
```
boxblinkracer, line 91: does it also work before 6.4? do you know that?

tests/Cypress/cypress/integration/storefront/payment-methods/creditcard.spec.js
```diff
@@ -157,6 +157,53 @@ describe('Credit Card Components', () => {
 
                 assertComponentErrors(true, true, true, false);
             })
+
+            it('Components are initialized after failed payment', () => {
```
boxblinkracer, line 161: can you make it less technical with something like
"Complete payment after failure with components" ...?

tests/Cypress/cypress/integration/storefront/payment-methods/creditcard.spec.js
```diff
@@ -157,6 +157,53 @@ describe('Credit Card Components', () => {
...
+                }
+
+                shopware.prepareDomainChange();
+                checkout.placeOrderOnConfirm();
+
+                cy.url().should('include', 'https://www.mollie.com/checkout/');
+
+                // verify that our component card is really
+                // been used by comparing the last 4 digits
+                cy.contains('**** ' + validCardNumber.substr(validCardNumber.length - 4));
```
boxblinkracer, line 182: can you remove this, its tested in another test
and then we have less steps in here

tests/Cypress/cypress/integration/storefront/payment-methods/creditcard.spec.js
```diff
@@ -157,6 +157,53 @@ describe('Credit Card Components', () => {
...
+                // have to close it in older versions
+                if (shopware.isVersionLower(6.4)) {
+                    payment.closePaymentsModal();
+                }
+
+                checkout.placeOrderOnEdit();
+
+                // Apparently we are not being sent to Mollie if we failed the first time and enter the same credentials again.
+                // It just immediately redirects to checkout/finish. Probably a bug in Shopware
+                cy.url().should('include', '/checkout/finish');
```
boxblinkracer, line 204: thats what you were talking about right?
so i would check debug this after the merge i guess?

boxblinkracer (changes requested):

tests/Cypress/cypress/integration/storefront/payment-methods/creditcard.spec.js
KienerNL: Would probably be better to do as a separate ticket

src/Resources/app/storefront/src/mollie-payments/plugins/creditcard-components-sw64.plugin.js
KienerNL: page layout, at least for the confirmOrderForm, has been the same since 6.2. It didn't exist at all in 6.1.5.

PR #242, opened 2021-11-29
MOL-222: Add PSR Logger for Mollie

boxblinkracer:
remove database event logging for now and add file logger 

the LoggerService (ex- event logger) has been rebuilt as MolliEventLogger
that one writes to the database, so we should use it wisely

the main logger is now psr based and logs to a separate file.
this one cant be seen in the admin but...yeah...whatever


src/Service/Logger/Services/IPAnonymizer.php
```diff
@@ -0,0 +1,56 @@
...
+     * @param string $ip
+     * @return string
+     */
+    public function anonymize($ip)
+    {
+        $ip = trim($ip);
+
+        # return an empty string and fail safe
+        # for our logger if the IP address is not even valid
+        if (!$this->isValidIP($ip)) {
```
schmitzcarsten, line 37: should the ip be anonymized always. Or should it be a setting for debug mode?

schmitzcarsten (approved):

src/Service/Logger/Services/IPAnonymizer.php
boxblinkracer: always :)

PR #245, opened 2021-12-02
MOL-654: Refund payments using ordernumber

KienerNL:
One new api endpoint has been added for refunding orders.

`/api/mollie/refund/order`
`[GET] number` ordernumber of the order to refund
`[GET] amount (optional)` amount to refund. 
Allows to refund an order using the ordernumber. 

If no amount is given, the amount will be determined based on the total order amount in Shopware, and the already refunded amount in Mollie. 
If any refunds are still pending in Mollie, a refund cannot be created without specifying an amount.


src/Controller/Api/RefundController.php
```diff
@@ -54,6 +56,32 @@ public function __construct(
...
+
+        if ($orderNumber === null) {
+            throw new \InvalidArgumentException('Missing Argument for Order Number!');
+        }
+
+        $order = $this->orderService->getOrderByNumber($orderNumber, $context);
+
+        $amount = (float)$query->get('amount', $order->getAmountTotal() - $this->refundService->getRefundedAmount($order, $context));
+
+        $description = $query->get('description', sprintf("Refunded through Shopware API. Order number %s",
```
boxblinkracer, line 79: PR all ready to merge
could you just add the PSR logger and add a INFO log in here?
...also with the text refunded order XY through Shopware API...i guess this might be good.
DI service is mollie_payments.logger I think, thank you

boxblinkracer (changes requested):
all good, just the log would be nice, thank you

PR #249, opened 2021-12-08
MOL-629

iamreinder:


src/Service/PaymentMethodService.php
```diff
@@ -156,41 +154,49 @@ public function addPaymentMethods(Context $context) : void
...
-                            'name' => $paymentMethod['description'],
-                        ];
-                    }
+        $installedHandlers = [];
+        $paymentCriteria = new Criteria();
+        $paymentCriteria->addFilter(new ContainsFilter('handlerIdentifier', 'MolliePayments'));
+
+        $paymentMethods = $this->paymentRepository->search($paymentCriteria, $context);
+
+        if ($paymentMethods->count()) {
```
boxblinkracer, line 169: can we reduce the lines of code and complexity a bit? the foreach doesnt do anything if count is 0, so either early return or just let it iterate 0 times :)

src/Service/PaymentMethodService.php
```diff
@@ -202,16 +208,16 @@ public function activatePaymentMethods(MollieApiClient $apiClient, Context $cont
...
+     * @param bool $active
+     * @param Context $context
      *
      * @return EntityWrittenContainerEvent
      */
     public function activatePaymentMethod(
         string $paymentMethodId,
-        bool $active = true,
-        Context $context = null
+        bool $active,
```
boxblinkracer, line 219: thanks for removing the default values :)

src/Service/PaymentMethodService.php
```diff
@@ -247,23 +254,49 @@ public function getPaymentMethodById($id) : ?PaymentMethodEntity
...
+     * @return array
+     */
+    public function getInstallablePaymentMethods() : array
+    {
+        // Variables
+        $paymentMethods = [];
+        $installablePaymentMethods = $this->getPaymentHandlers();
+
+        // Add payment methods to array
+        if ($installablePaymentMethods !== null) {
```
boxblinkracer, line 269: early return please

tests/PHPUnit/Service/PaymentMethodServiceTest.php
```diff
@@ -0,0 +1,272 @@
...
+        $this->setUpPaymentMethodUpsertReturn();
+
+        $installablePaymentMethods = $this->paymentMethodService->getInstallablePaymentMethods();
+        $installablePaymentMethod = array_shift($installablePaymentMethods); // we pick one, so we don't need to fake a large number of results
+
+        $this->paymentMethodService->addPaymentMethods([$installablePaymentMethod], $this->context);
+
+        $upsertResult = $this->paymentMethodRepository->data;
+
+        self::assertNotEmpty(
```
boxblinkracer, line 196: question, these are nice asserts, but if its empty the assert below would already be failing right? so we can reduce the lines of codes and visual complexity for this test and still keep the same results by removing those additional notEmpty assertions? if so, could you do that? :)

tests/PHPUnit/Service/PaymentMethodServiceTest.php
```diff
@@ -0,0 +1,272 @@
...
+            $upsertResult,
+            'The upserted data from method addPaymentMethods is expected not to be empty when providing installable payment methods.'
+        );
+
+        self::assertSame(
```
boxblinkracer, line 201: i might seem a bit stupid, but i'm wondering what is tested in here, is the actual either $upsertResult or $installablePaymentMethod
according to the text in L203 its $upserResult. but the first argument in the assert is always the "expected" as far as i know. so could it be that this is inverted?
maybe you could rename the expected variable to something like $expectedPaymentMethods :) that could also help

boxblinkracer (changes requested):
see comments thank you

src/Service/PaymentMethodService.php
iamreinder: 👍🏻

src/Service/PaymentMethodService.php
iamreinder: I will implement a fast return 👌🏻

src/Service/PaymentMethodService.php
iamreinder: ✅

tests/PHPUnit/Service/PaymentMethodServiceTest.php
iamreinder: Ah yes, I did reverse them. Sorry. I will make it more easily understandable.

tests/PHPUnit/Service/PaymentMethodServiceTest.php
iamreinder: Ah yes, I did reverse them. Sorry. I will make it more easily understandable.

PR #250, opened 2021-12-09
MOL-603: Chargeback status in Shopware 6

KienerNL:
Adds transition for chargeback status in Shopware 6. This transition will be triggered when the payment has an "amountChargedBack" greater than 0.

Note: 
- The chargeback state and action did not exist in Shopware 6.2.2 and lower, payment status will be set to in progress instead.
- Testing chargebacks is presently not yet possible with Mollie in test mode.


src/Compatibility/Gateway/CompatibilityGateway.php
```diff
@@ -73,6 +75,41 @@ public function getLineItemPromotionType(): string
         return 'promotion';
     }
 
+    /**
+     * @return string
+     */
+    public function getChargebackOrderTransactionState(): string
+    {
+        // Chargeback state did not exist before 6.2.3, so set to in progress instead.
+        if(!$this->versionGTE('6.2.3')) {
```
boxblinkracer, line 84: could you add a LT....i read if its greater than and was wondering
i think it might exist, otherwise feel free to add it

src/Resources/config/config.xml
```diff
@@ -248,5 +248,43 @@
                 </option>
             </options>
         </input-field>
+        <input-field type="single-select">
+            <name>orderStateWithAChargebackTransaction</name>
+            <label>Order state with a charged back transaction</label>
+            <label lang="de-DE">Bestellstatus bei einer wiederhergestellten Transaktion</label>
```
boxblinkracer, line 254: wiederhergestellt should more be "einer Rückbuchung" or "einer zurückgebuchten Transaktion" (more in the transaction-pattern), thats what "teruggevorderde" would be in german

src/Service/Mollie/OrderStatusConverter.php
```diff
@@ -75,6 +81,10 @@ public function getMolliePaymentStatus(?Payment $payment = null): string
             $status = MolliePaymentStatus::MOLLIE_PAYMENT_EXPIRED;
         }
 
+        if($payment->amountChargedBack && $payment->amountChargedBack->value > 0) {
+            $status = MolliePaymentStatus::MOLLIE_PAYMENT_CHARGEBACK;
```
boxblinkracer, line 85: attention! i think its still paid, so that has to be done BEFORE isPaid...

boxblinkracer: ah, screw that, thought its an early return :D 
but maybe you can change it to early return? im a bit scared someone builds and early return one day and thinks thats correct.
maybe you can even add it inside the isPaid IF? then it would be clearly visible that its only in there..just like refunded which seems to be missing in here?! :D 

now that i see it, could it be that the order status function has the same payment-status check but is missing, so duplicate code, but refund is missing?
if so we should clean and fix it...but maybe in a separate PR?

boxblinkracer (changes requested):
see comments, thank you

src/Service/Mollie/OrderStatusConverter.php
KienerNL: Should it not always be set to chargeback state, if even a tiny amount has been charged back? Disregarding any other status it had before.

Refund status is done in the `getMollieOrderStatus` method, using the amounts on the order object, not on the payment. This should indeed also be added to the `getMolliePaymentStatus` method.

If `getMolliePaymentStatus` returns "unknown", either because it does not have a Payment (which would be weird) or the Payment has a status we're not checking for, `getMollieOrderStatus` will do a similar check on its status methods.

PR #253, opened 2021-12-14
MOL-223 update payment methods button in administration

iamreinder:


src/Resources/app/administration/src/module/mollie-payments/components/mollie-update-payment-methods/mollie-update-payment-methods.html.twig
```diff
@@ -0,0 +1,6 @@
+<sw-container gap="16px" columns="1fr" justify="start">
+    <sw-button @click="onUpdatePaymentMethodsButtonClicked" ref="updatePaymentMethodsButton">
+        <sw-loader v-if="updatePaymentMethodsIsLoading" size="16px"></sw-loader>
+        {{ $tc('sw-payment.updatePaymentMethods.button') }}
```
boxblinkracer, line 4: could you put that in the sub node mollie_payments (i think thats the name)

iamreinder: Fixed it 🙂

src/Controller/Api/PaymentMethodController.php
```diff
@@ -0,0 +1,51 @@
...
+     * @return JsonResponse
+     */
+    public function updatePaymentMethods(Context $context): JsonResponse
+    {
+        $success = true;
+
+        try {
+            $this->paymentMethodService->installAndActivatePaymentMethods($context);
+        } catch (Exception $e) {
+            $success = false;
```
boxblinkracer, line 44: i think we really want a logger in here?

also, what happens if you then throw the exception again? does it also lead to an error in the admin?
then we can make it fail fast
would it expose confidental data? then keep it as it is,

src/Resources/app/administration/src/core/service/api/mollie-payments-payment-method.service.js
```diff
@@ -0,0 +1,25 @@
+// eslint-disable-next-line no-undef
+const ApiService = Shopware.Classes.ApiService;
+
+class MolliePaymentsPaymentMethodService extends ApiService {
+    constructor(httpClient, loginService, apiEndpoint = 'mollie') {
```
boxblinkracer, line 5: love that one, thanks for doing OOP stuff

boxblinkracer (changes requested):
see comments

src/Controller/Api/PaymentMethodController.php
iamreinder: I'm not sure. It could both lead to database errors of Shopware, as well as API exceptions from Mollie. I don't exactly know which errors could be thrown and if there might be confidential data in there.

On the other hand, the user has access to the plugin configuration if it triggers this method. That wouldn't be someone who shouldn't have access to the data, right?

src/Resources/app/administration/src/core/service/api/mollie-payments-payment-method.service.js
iamreinder: 🙂 👍🏻

src/Controller/Api/PaymentMethodController.php
iamreinder: Correction: it could not lead to API exceptions from Mollie. It doesn't do anything with Mollie. It just works with the handler classes. So it would only be database errors of Shopware. I guess those are safe to show, right?

PR #257, opened 2021-12-21
MOL-678 fixed duplicate payment methods after renaming and updating

iamreinder:
Built on feature [MOL-223](https://github.com/mollie/Shopware6/pull/253) because it uses the same services. So please, first merge [MOL-223](https://github.com/mollie/Shopware6/pull/253) before reviewing this bugfix.


schmitzcarsten (approved):

src/Controller/Api/PaymentMethodController.php
```diff
@@ -0,0 +1,51 @@
...
+     * @return JsonResponse
+     */
+    public function updatePaymentMethods(Context $context): JsonResponse
+    {
+        $success = true;
+
+        try {
+            $this->paymentMethodService->installAndActivatePaymentMethods($context);
+        } catch (Exception $e) {
+            $success = false;
```
boxblinkracer, line 44: can you please add a log so we know that something did not work please?
also if that is an internal api, then please also add the error for the admin if possible.

i guess you can catch the exception to add a logger entry, and then pass it on and re-throw it to also see it automatically by the admin? if not then I'm fine, but we need the log please

src/Resources/app/administration/src/module/mollie-payments/components/mollie-update-payment-methods/index.js
```diff
@@ -0,0 +1,73 @@
...
+                .then((response) => {
+                    const messageData = {
+                        title: me.$tc('mollie-payments.config.payments.updatePaymentMethods.title'),
+                        message: me.$tc('mollie-payments.config.payments.updatePaymentMethods.succeeded'),
+                    };
+
+                    if (response.success === true) {
+                        me.createNotificationSuccess(messageData);
+                    } else {
+                        messageData.message = me.$tc('mollie-payments.config.payments.updatePaymentMethods.failed');
```
boxblinkracer, line 39: ok i see, here is the message.
so depending on if you add it to the response you could show it here, 
the logs cannot be seen in the admin, so it might be a cool thing to see it in the alert if possible.
then things like "wrong api key" might not even reach us at support haha :)

what do you think?

boxblinkracer (changes requested):
great work, just a small addition to the error handling and visible messages :)

src/Resources/app/administration/src/module/mollie-payments/components/mollie-update-payment-methods/index.js
iamreinder: This is actually part of MOL-223 (#253). There is no communication with the Mollie API at this point. It's just the payment handlers being installed. As we agreed upon, the payment methods can be safely activated here (unless a payment method already exists and is deactivated) because a merchant still has to assign them in the storefront.

src/Controller/Api/PaymentMethodController.php
iamreinder: This was part of MOL-233 (#253), solved in that issue, now pulled into this branch.

PR #260, opened 2021-12-23
MOL-645: Fix and rebuild partial shipping

KienerNL:
Partial shipping in the administration has been rebuilt.

In addition, a new button has been added to allow shipping the entire order. You can find it in the new dropdown menu under the "Refund with Mollie" button.

Furthermore, support for tracking codes has been added to all shipping functions:

- Both the "Ship Order" and "Ship Item" functionalities in the administration allow you to enter tracking info. ¹
- The "human-readable" shipping API endpoints added previously now also support tracking info. ²
- An event has been added that can be used by shipment plugins (Sendcloud, PostNL, etc) to automatically ship the order and add their tracking information to Mollie. If the Shopware order was not paid with Mollie, it will do nothing of course. ³

**Tracking info**

1. If you choose to add tracking info to Mollie, then both a carrier and a tracking code need to be entered. A tracking url is optional, but nice to have.
2. The shipping API endpoints now support `trackingCarrier`, `trackingCode` and `trackingUrl` parameters. As with ¹, `trackingCarrier` and `trackingCode` are required if you specify either of them. `trackingUrl` is completely optional.
3. To add your tracking info to and ship the order with Mollie, dispatch the new `MollieOrderShipmentTrackingEvent` using the `event_dispatcher` service. As with ¹ and ² `trackingCarrier` and `trackingCode` are required, `trackingUrl` is optional. In addition, the `orderId` of the order to ship and add tracking info to is required.


src/Subscriber/ShippingSubscriber.php
```diff
@@ -0,0 +1,63 @@
...
+                $event->getOrderId(),
+                $event->getTrackingCarrier(),
+                $event->getTrackingCode(),
+                $event->getTrackingUrl(),
+                $event->getContext()
+            );
+        } catch (CouldNotExtractMollieOrderIdException $e) {
+            // We need to catch CouldNotExtractMollieOrderIdException, because if it's not a Mollie Order
+            // it obviously cannot get shipped with Mollie. We also don't have to log this.
+            // But if we don't catch it, the rest of the process might break.
```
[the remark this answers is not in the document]
KienerNL: Good idea, I should really put debug logs in more places.

src/Event/MollieOrderShipmentTrackingEvent.php
```diff
@@ -0,0 +1,129 @@
...
+     */
+    protected $trackingUrl = '';
+
+
+    public function __construct(
+        string  $orderId,
+        Context $context,
+        string  $trackingCarrier,
+        string  $trackingCode,
+        string  $trackingUrl = ''
```
boxblinkracer, line 40: please no optional arguments. lets stick with hard definitions

src/Facade/MollieRefundFacade.php
```diff
@@ -68,23 +68,22 @@ public function refundUsingOrderNumber(string $orderNumber, float $amount, strin
...
      * @param string $description
-     * @param Context $context
      * @return Refund
      */
-    private function refund(OrderEntity $order, float $amount, string $description, Context $context): Refund
+    private function refund(OrderEntity $order, float $amount, string $description): Refund
     {
         if ($amount === 0.0) {
-            $amount = $order->getAmountTotal() - $this->refundService->getRefundedAmount($order, $context);
+            $amount = $order->getAmountTotal() - $this->refundService->getRefundedAmount($order);
```
boxblinkracer, line 83: is this a helper to have a refundAll or refundAmount?
because what i dont like is, that an accidental calling of 0 leads to a full refund.
meaning, if somewhere a cast goes wrong and a float cast leads to 0, then the full order is refunded :D 

maybe you can split the cases into 2 functions like refundAmount or refundFull? 
> IF i understand it correctly

src/Facade/MollieShipment.php
```diff
@@ -281,4 +444,25 @@ private function findMatchingLineItems(OrderEntity $order, string $itemIdentifie
...
         });
     }
+
+    private function createTrackingInfoStruct(
+        string $trackingCarrier,
+        string $trackingCode,
+        string $trackingUrl
+    ): ?ShipmentTrackingInfoStruct
+    {
+        if(empty($trackingCarrier) && empty($trackingCode)) {
```
boxblinkracer, line 454: great conditions
would be great to have that in a unit test ;)

src/Service/MollieApi/Shipment.php
```diff
@@ -98,4 +108,69 @@ public function shipItem(
...
+                continue;
+            }
+
+            $orderLineItemId = $mollieOrderLine->metadata->orderLineItemId ?? null;
+            if (empty($orderLineItemId)) {
+                continue;
+            }
+
+            $lineItems[$orderLineItemId] = [
+                'id' => $orderLineItemId,
```
boxblinkracer, line 135: can we do a unit test for that array and the skippings above?

src/Service/MollieApi/Shipment.php
```diff
@@ -98,4 +108,69 @@ public function shipItem(
...
+    public function getTotals(string $mollieOrderId, string $salesChannelId): array
+    {
+        $mollieOrder = $this->orderApiService->getMollieOrder($mollieOrderId, $salesChannelId);
+
+        $totalAmount = 0.0;
+        $totalQuantity = 0;
+
+        foreach ($mollieOrder->lines() as $mollieOrderLine) {
+            /** @var OrderLine $mollieOrderLine */
+            if ($mollieOrderLine->type === OrderLineType::TYPE_SHIPPING_FEE) {
```
boxblinkracer, line 160: maybe here too

src/Struct/MollieApi/ShipmentTrackingInfoStruct.php
```diff
@@ -0,0 +1,87 @@
...
+    {
+        $this->carrier = $carrier;
+        $this->code = $code;
+        $this->url = $url;
+    }
+
+    /**
+     * @return array<string, string>
+     */
+    public function toArray() {
```
boxblinkracer, line 32: could you add unit tests for this please?
perfect thing to verify the structure,
like the idea of this struct

src/Subscriber/ShippingSubscriber.php
```diff
@@ -0,0 +1,69 @@
...
+    )
+    {
+        $this->shipmentFacade = $shipmentFacade;
+        $this->logger = $logger;
+    }
+
+    public static function getSubscribedEvents()
+    {
+        return [
+            MollieOrderShipmentTrackingEvent::class => 'onShipOrderWithTracking',
```
boxblinkracer, line 35: amazing idea.
1 thing to figure out...would a different plugin create an object of type MollieOrderShipmentTrackingEvent?
if so, then ..yeah..please switch over to an array. otherwise the shop would crash, if one decides to disable or uninstall the mollie plugin but keep PostNL or whatever plugin it uses.
i know it still would crash if one would subscribe to ::class, but if he uses a decoupled string, then its completely loose and would not crash

what do you think?

src/Subscriber/ShippingSubscriber.php
```diff
@@ -0,0 +1,69 @@
...
+            // But if we don't catch it, the rest of the process might break.
+            $this->logger->debug($e->getMessage(), [
+                'orderId' => $event->getOrderId(),
+                'trackingCarrier' => $event->getTrackingCarrier(),
+                'trackingCode' => $event->getTrackingCode(),
+                'trackingUrl' => $event->getTrackingUrl(),
+            ]);
+        } catch (\Exception $e) {
+            // We log the error, but don't rethrow so the rest of the proces can continue.
+            $this->logger->error($e->getMessage(), [
```
boxblinkracer, line 61: i dont know what is in there, but maybe add a text line before like
"Error when shipping order from Mollie Event" or something like this to locate the error also on a text-level besides introspection? so like semantic over syntax :)

boxblinkracer (changes requested):
please see comments

src/Subscriber/ShippingSubscriber.php
KienerNL: I would hope they'd use a `class_exists` before trying to implement it, instead of just blindly assuming it exists :smile: 

What do you mean by switching it to an array?

src/Facade/MollieRefundFacade.php
KienerNL: Perhaps it'd be better to do this in a separate ticket, as it's already in master at the moment and this is just removing an unused parameter.

PR #262, opened 2022-01-04
NTR: Add transaction id to custom fields

fjbender:
As a merchant doing my payments reconciliation
I want the Mollie payment ID per shop order
because it is listed in the payout CSV reports.

A line in a Mollie settlement file might look like:

```csv
"2021-12-07 08:25:56",klarnapaylater,EUR,129.00,,cpt_fooacr,"Capture of Pay later. transaction on 06-12-2021 to Max Merchant GmbH. Original payment: tr_xfOoBK8e8P - 'Bestellung 10083'","Constantin Consumer",,,EUR,129.00,123456789.2112.03,
```

In the order's custom fields we only log the `ord_` ID, which isn't reported back during reconciliation.

The merchant might try to match outpayments and orders by the order number part of the description, which can work, but using the `tr_` ID might be easier.

PayPal adds their ID to the transaction instead of the order, so maybe it might be worth the while to also do that: https://github.com/shopwareLabs/SwagPayPal/blob/4543ae7896cee34275f8fac397ef52431f838f65/src/Checkout/Payment/Handler/AbstractPaymentHandler.php#L25-L40

### DIFF
--- a/src/Facade/MolliePaymentFinalize.php
+++ b/src/Facade/MolliePaymentFinalize.php
@@ -9,6 +9,7 @@ use Kiener\MolliePayments\Service\Mollie\OrderStatusConverter;
 use Kiener\MolliePayments\Service\Order\OrderStatusUpdater;
 use Kiener\MolliePayments\Service\SettingsService;
 use Kiener\MolliePayments\Service\Transition\TransactionTransitionServiceInterface;
+use Kiener\MolliePayments\Service\UpdateOrderCustomFields;
 use Kiener\MolliePayments\Struct\MollieOrderCustomFieldsStruct;
 use Mollie\Api\Exceptions\ApiException;
 use Mollie\Api\Exceptions\IncompatiblePlatform;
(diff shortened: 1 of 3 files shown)


src/Facade/MolliePaymentFinalize.php
```diff
@@ -77,6 +84,12 @@ public function finalize(AsyncPaymentTransactionStruct $transactionStruct, Sales
         $apiClient = $this->mollieApiFactory->getClient($salesChannelContext->getSalesChannel()->getId());
         $mollieOrder = $apiClient->orders->get($mollieOrderId, ['embed' => 'payments']);
 
+        // Add the transaction ID to the order's custom fields
+        // We might need this later on for reconciliation
+        $molliePaymentId = $mollieOrder->_embedded->payments[0]->id;
```
fjbender, line 89: Not sure here if this is the smartest way to access an embedded payment.

src/Facade/MolliePaymentFinalize.php
```diff
@@ -76,6 +91,27 @@ public function finalize(AsyncPaymentTransactionStruct $transactionStruct, Sales
 
         $apiClient = $this->mollieApiFactory->getClient($salesChannelContext->getSalesChannel()->getId());
         $mollieOrder = $apiClient->orders->get($mollieOrderId, ['embed' => 'payments']);
+        $molliePaymentId = end($mollieOrder->_embedded->payments)->id;
```
schmitzcarsten, line 94: Are the payments chronological ? I think they are not . Perhaps better order the array chronological and afterwards take the correct payment

schmitzcarsten: I think we had a service that fetches the last payment or the correct one. It's really critical, so use an existing service or extract it to a service and create unit tests

schmitzcarsten (changes requested):

src/Facade/MolliePaymentFinalize.php
fjbender: We do it like this in the magento 2 module as well, but I get that it's not ideal. There is a service that does that (`\Kiener\MolliePayments\Service\MollieApi\Order::getCompletedPayment`), but it would mean to do a second API call like the one done before. The other ways would be to either duplicate the code or refactor this entire part of the code. I'll see what I can come up with, thanks!

fjbender: Opted for the two API call option for now.

PR #268, opened 2022-01-13
Fix payment specific params not being added to payment on reattempt

KienerNL:
Payment specific parameters were not passed along to Mollie when creating a new payment after the first one failed.

### DIFF
--- a/src/Resources/config/services.xml
+++ b/src/Resources/config/services.xml
@@ -45,7 +45,7 @@
             <argument type="service" id="customer.repository"/>
             <argument type="service" id="Kiener\MolliePayments\Service\MollieApi\Customer"/>
             <argument type="service" id="event_dispatcher"/>
-            <argument type="service" id="monolog.logger"/>
+            <argument type="service" id="mollie_payments.logger"/>
             <argument type="service" id="Shopware\Core\System\SalesChannel\Context\SalesChannelContextPersister"/>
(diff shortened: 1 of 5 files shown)


src/Resources/config/services.xml
```diff
@@ -45,7 +45,7 @@
             <argument type="service" id="customer.repository"/>
             <argument type="service" id="Kiener\MolliePayments\Service\MollieApi\Customer"/>
             <argument type="service" id="event_dispatcher"/>
-            <argument type="service" id="monolog.logger"/>
+            <argument type="service" id="mollie_payments.logger"/>
```
boxblinkracer, line 48: huch, perfect thanks :D

src/Service/MollieApi/Order.php
```diff
@@ -210,6 +220,41 @@ private function getOpenPayment(MollieOrder $mollieOrder): ?Payment
...
+    private function prepareNewPayment(
+        MollieOrder    $mollieOrder,
+        string         $paymentMethod,
+        PaymentHandler $paymentHandler,
+        OrderEntity    $order,
+        CustomerEntity $customer,
+        SalesChannelContext $salesChannelContext
+    ): Payment
+    {
+        $fakeOrder = [
```
boxblinkracer, line 242: could you just add a few lines why this is "fake" please

boxblinkracer (approved):

PR #295, opened 2022-02-17
MOL-745: Do not change description when updating payment methods

KienerNL:


src/Service/PaymentMethodService.php
```diff
@@ -149,6 +149,8 @@ public function addPaymentMethods(array $paymentMethods, Context $context) : voi
             if (isset($existingPaymentMethod)) {
                 $paymentMethodData['id'] = $existingPaymentMethod->getId();
                 $paymentMethodData['name'] = $existingPaymentMethod->getName();
+                $paymentMethodData['description'] = $existingPaymentMethod->getDescription();
```
boxblinkracer, line 152: perfect thanks
might there be something else too that needs to be added from the old method?
i'm thinking of when custom data might be existing?
shouldnt we just use the old data and only overwrite what should be changed...also...currently thinking....why are we actually updating anything at all? :D 

just a few thoughts...if not correct, then I would just merge it. code seems good
thanks

PR #297, opened 2022-02-17
Add MollieOrderBuildEvent

runelaenen:
This PR adds an event that allows developers to hook into the order builder.

### DIFF
--- a/src/Event/MollieOrderBuildEvent.php
+++ b/src/Event/MollieOrderBuildEvent.php
@@ -1,0 +1,108 @@
+<?php
+
+namespace Kiener\MolliePayments\Event;
+
+use Shopware\Core\Checkout\Order\OrderEntity;
+use Shopware\Core\System\SalesChannel\SalesChannelContext;
+
+class MollieOrderBuildEvent
+{
+    /**
+     * @var array
+     */
+    private $orderData;
+
+    /**
+     * @var OrderEntity
+     */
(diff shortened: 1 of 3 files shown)


boxblinkracer:
Hi @runelaenen 
thank you for this

can you tell me why we need this?
so a use case...

This has indeed been discussed...problem is...we have to guarantee a bit that payments work 😅  ...
and people could mess up things with this 😄

runelaenen:
The 36 is the listing price, and the 24 is the normal price.
This is filled in the cart collector by a custom script, and not by Shopware - which is probably where it goes wrong somewhere. However, this is the first place where we notice it not working as expected. The cart shows it correctly as a discounted price.

It would save me a lot of headaches if I had an event here where I can fix the pricing for these custom discounted line items.

boxblinkracer:
Hi @runelaenen 

ah, now it makes sense....because honestly...it seemed as if there is a bug in your custom script.
this feels a bit better now, otherwise i would have been really....really confused

besides that
I totally see the point, let me check with Mollie and do some internal "risk management" :D
please keep the PR open, i might merge it

thank you for your work on this

boxblinkracer:
hi
thank you for this
i did a separate PR with some changes
the final decision was to start by only allowing custom meta data strings, let's see where this is going and then we might open more possibilities

besides that, amazing idea, thank you so much for this

PR #299, opened 2022-02-21
MOL-657: Refund warnings #3

KienerNL:


boxblinkracer:
attention, it looks like this,
i had that in the first merge, so we should rather fix it directly in the PR this time 

Screen Shot 2022-02-22 at 13 43 58

boxblinkracer:
feature will be closed, we're doing something else

PR #311, opened 2022-03-03
MOL-746: JS error with iDeal issuers

daniel-memo-ict:
This occurs when
- iDeal is activated
- Mollie Limits are disabled
- An invalid API key is being used.

In this case iDeal will remain visible but we can't get available issuers from Mollie due to the invalid API key, and thus no issuers dropdown is available to register events on. If that's the case, skip rest of the plugin execution as everything hinges on the dropdown.

### DIFF
--- a/src/Subscriber/SystemConfigSubscriber.php
+++ b/src/Subscriber/SystemConfigSubscriber.php
@@ -6,6 +6,7 @@ use Kiener\MolliePayments\Helper\ProfileHelper;
 use Kiener\MolliePayments\Service\SettingsService;
 use Kiener\MolliePayments\Setting\MollieSettingStruct;
 use Mollie\Api\MollieApiClient;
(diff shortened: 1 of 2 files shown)


src/Subscriber/SystemConfigSubscriber.php
```diff
@@ -136,6 +137,19 @@ private function fetchProfileIdForApiKey($value, ?string $salesChannelId, bool $
         $this->apiClient->setApiKey($value);
 
         $profile = ProfileHelper::getProfile($this->apiClient, new MollieSettingStruct());
+
+        if(!$profile instanceof Profile) {
+            $this->logger->error(
+                'Could not get profile using these settings',
+                [
+                    'apiKey' => $value,
```
boxblinkracer, line 145: please no api key logging!

boxblinkracer (changes requested):
see comments

PR #337, opened 2022-04-13
MOL-780: Mollie Limits must not be used in the Account area

smemoict:
The easiest way to fix this was loading the data again if removing resulted in empty data.


src/Compatibility/Storefront/Route/PaymentMethodRoute/MollieLimits/MollieLimitsPaymentMethodRoute64.php
```diff
@@ -71,8 +63,12 @@ public function getDecorated(): AbstractPaymentMethodRoute
     public function load(Request $request, SalesChannelContext $context, Criteria $criteria): PaymentMethodRouteResponse
     {
         $originalData = $this->corePaymentMethodRoute->load($request, $context, $criteria);
-
-        return $this->mollieLimits->removePaymentMethods($originalData, $context);
+        $newData = $this->mollieLimits->removePaymentMethods($originalData, $context);
+        
+        if (count($newData->getPaymentMethods()) == 0) {
```
boxblinkracer, line 68: i dont think this work.
if i have a klarna plugin installed, then the count is > 0, right?

PR #339, opened 2022-04-15
MOL-780: Limits must not be used in the Account area (and footer)

daniel-memo-ict:
Can close #337


daniel-memo-ict:
Why is the 6.3 payment method route not defined anywhere in the container/dependency injection?

PR #356, opened 2022-06-01
Save reset stock in order line item custom field with refund

MelvinAchterhuis:
This PR makes it possible to keep track of the reset stock in Shopware through the API.

Screenshot 2022-06-01 at 15 54 25

When there are multiple refunds for a line item the resetStock gets added up.


MelvinAchterhuis:
Hi @boxblinkracer ,

Do you have an example with the "attribute" class? I will implement it + fix the failing test :)

boxblinkracer:
hi
yes, like this
https://github.com/mollie/Shopware6/blob/master/src/Struct/Product/ProductAttributes.php
then one could simply say

```ruby
$prodAtts = new ProductAttributes($productEntity);
$type = $prodAtts->getVoucherType();
```

MelvinAchterhuis:
@boxblinkracer , could you have another look at the PR?

PR #365, opened 2022-06-23
MOL-784-iDeal-with-default-dropdown-"Select-Bank"

smemoict:
Validation is being triggered but no message is shown. I have no idea why that is not happening.


boxblinkracer:
hi thanks for this
you added the select as required...so its required to have a value
but the snippet text isnt linked to anything right? what do you want to display as mesage? what would you expect?

smemoict:
Normally you would get a browser native message like in this [example](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/required)

boxblinkracer:
hi, a alright, sry my mistake, got you
i can check it out tomorrow and see if it works on my end.

boxblinkracer:
hi
alright, so this works good,
i figured out what you mean, its just not red when its not valid, ill just add the css for that :)
so ....time for a merge,...great work :)

PR #366, opened 2022-06-23
MOL-675-Show-Payment-Method-Transaction-Reference-in-Administration

smemoict:
I don't know what kind of label would be preferred here so I added some to the best of my knowledge.


boxblinkracer:
nice one, thanks :), ill check it out and in case the look doesnt fit, ill add it, but i guess its fine

PR #371, opened 2022-06-30
NTR: payment methods are not removed on order edit

daniel-memo-ict:
This refactors the 3 payment method routes somewhat, and abstracts some functions into a base payment method remover class, as well as providing interfaces.


src/Compatibility/Storefront/Route/PaymentMethodRoute/Subscription/SubscriptionPaymentMethodRoute64.php
```diff
@@ -53,17 +44,19 @@ public function getDecorated(): AbstractPaymentMethodRoute
...
+     * @param Criteria            $criteria
      * @return PaymentMethodRouteResponse
      * @throws \Exception
      */
     public function load(Request $request, SalesChannelContext $context, Criteria $criteria): PaymentMethodRouteResponse
     {
         $originalData = $this->corePaymentMethodRoute->load($request, $context, $criteria);
 
-        return $this->subscriptionRemover->removePaymentMethods($originalData, $context);
+        $newData = $this->paymentMethodRemover->removePaymentMethods($originalData, $context);
```
boxblinkracer, line 57: is that duplicated?

daniel-memo-ict: *waves hand* you saw nothing

PR #377, opened 2022-07-18
Mol 627: Save credit card issuer in shopware order, and display it.

smemoict:
I've added credit card data to the order custom data and components to display it in the admin. (With an icon!)


boxblinkracer:
Hi

this is how it looks for me? I don't see anything (NGROK is also used)

10003___Orders___Shopware_Administration

smemoict:
Hey, there was indeed an error of cc data not being saved on account of being overwritten. This has been remedied and should work now.

boxblinkracer:
awesome
thank you @smemoict 
ill check it out tomorrow :)

src/Struct/Order/OrderAttributes.php
```diff
@@ -46,6 +44,42 @@ class OrderAttributes
      */
     private $molliePaymentUrl;
 
+    /**
+     * @var null|string
+     */
+    private $creditCardNumber;
```
boxblinkracer, line 50: can we get red if the null|string please
always make sure to return a string, if null, then its just empty
that helps for all developers that consume the data, you can just check if not empty instead of also do the NULL check.
only use null if really 3 states exist like "not defined", "empty" and "filled".
in this case its just not existing, or existing :) 
that reduces risks :)

src/Service/OrderService.php
```diff
@@ -226,6 +227,11 @@ public function updateMollieDataCustomFields(OrderEntity $order, string $mollieO
             if (isset($molliePayment->details, $molliePayment->details->transferReference)) {
                 $thirdPartyPaymentId = $molliePayment->details->transferReference;
             }
+
+            # check for creditcard
+            if (isset($molliePayment->method, $molliePayment->details) && $molliePayment->method === "creditcard") {
```
boxblinkracer, line 232: please use constant PaymentMethod::CREDITCARD

src/Resources/app/administration/package.json
```diff
@@ -1,30 +1,31 @@
...
-        "eslint-plugin-node": "^11.1.0",
-        "eslint-plugin-promise": "^5.1.0",
-        "eslint-plugin-standard": "^5.0.0",
-        "flow-bin": "^0.160.0",
-        "jest": "^27.1.1",
-        "stylelint": "^14.1.0",
-        "stylelint-config-standard": "^24.0.0",
-        "stylelint-scss": "^4.0.1"
-    }
+  "name": "mollie-shopware6-administration-tests",
```
boxblinkracer, line 2: not quite sure why this file is touched? i see the raw-loader
you also added a webpack config, so isnt that then a prod-dependency and not a dev one? this file was originally meant for the testing
...im also wondering why its needed...i think there are already images existing in the administration...so i dont get why this has been added :) maybe you can explain

boxblinkracer (changes requested):
its now working on my side
thank you :)
started code review and found a few small things, please see comments

src/Resources/app/administration/package.json
smemoict: This was a formatting commit that slipped though, I think I reverted it.

src/Service/OrderService.php
smemoict: agreed, done

src/Struct/Order/OrderAttributes.php
smemoict: Sounds good

PR #388, opened 2022-07-29
MOL-859: The 'streetAdditional' field should not contain only whitespace characters

daniel-memo-ict:
Trimmed the value so whitespace characters at the start and end are removed. If the string ends up with a length of 0, send null instead

### DIFF
--- a/src/Service/MollieApi/Builder/MollieOrderAddressBuilder.php
+++ b/src/Service/MollieApi/Builder/MollieOrderAddressBuilder.php
@@ -20,7 +20,7 @@ class MollieOrderAddressBuilder
             'familyName' => $address->getLastName(),
             'email' => $email,
             'streetAndNumber' => $address->getStreet(),
-            'streetAdditional' => $address->getAdditionalAddressLine1(),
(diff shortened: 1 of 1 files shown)


src/Service/MollieApi/Builder/MollieOrderAddressBuilder.php
```diff
@@ -20,7 +20,7 @@ public function build(string $email, ?CustomerAddressEntity $address): array
             'familyName' => $address->getLastName(),
             'email' => $email,
             'streetAndNumber' => $address->getStreet(),
-            'streetAdditional' => $address->getAdditionalAddressLine1(),
+            'streetAdditional' => trim($address->getAdditionalAddressLine1()) ?? null,
```
boxblinkracer, line 23: street must not exist if empty or only whitespaces
so we need an if-statement that adds it if existing please

boxblinkracer (changes requested):
see comments please

boxblinkracer:
Hi
sorry i needed it for the new release, so i had to do it in https://github.com/mollie/Shopware6/pull/397
i will now close this PR

PR #389, opened 2022-07-29
MOL-768: plugin config throws console err

daniel-memo-ict:


src/Resources/app/administration/src/core/service/utils/version-compare.utils.js
```diff
@@ -0,0 +1,138 @@
+export default {
+    equals(versionA, versionB) {
```
boxblinkracer, line 2: hi
I figured out that my sales channel list was not displayed, even though its a 6.4.14.0.
then i discovered that the comparing does not completely work.

i've added this jest test and it shows that it fails.
can you fix it and add jest tests for it please?

```
const VersionCompare = require('../../../src/core/service/utils/version-compare.utils').default;

test('GTE Comparing works', () => {
    const result = VersionCompare.greaterOrEqual('6.4.14.0', '6.4.2');
    expect(result).toBe(true);
});
```

boxblinkracer (changes requested):
please see comments, thank you

PR #404, opened 2022-08-19
MOL-619: CustomerService should use compatibility gateway

daniel-memo-ict:


daniel-memo-ict:
@boxblinkracer Why does it keep breaking during plugin install? It seems to trip over Shopware's own classes.

daniel-memo-ict:
@boxblinkracer Is this still even necessary now that the minimum version of the plugin is 6.3.5.2?

PR #405, opened 2022-08-19
MOL-863: Subscription flow builder events should implement MailAware

daniel-memo-ict:


daniel-memo-ict:
commit https://github.com/mollie/Shopware6/pull/405/commits/5d271bc98249f8670b15ccce223e65a253b84ff5 might have actually been for MOL-870. I just did all the events in one go while I was working on them :smiley:

PR #416, opened 2022-09-02
Added payment method api route to valid cart routes

daniel-memo-ict:
Please also try this, should work for all payment method removers then. with only exception if we somehow dont have a cart object available.

### DIFF
--- a/src/Components/MollieLimits/Service/MollieLimitsRemover.php
+++ b/src/Components/MollieLimits/Service/MollieLimitsRemover.php
@@ -62,16 +62,19 @@ class MollieLimitsRemover extends PaymentMethodRemover
             return $originalData;
         }
 
-        try {
-            $cart = $this->getCart($context);
-        } catch (MissingCartServiceException $e) {
-            $this->logger->error($e->getMessage(), [
-                'exception' => $e,
-            ]);
-            return $originalData;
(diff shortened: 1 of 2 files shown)


src/Components/MollieLimits/Service/MollieLimitsRemover.php
```diff
@@ -62,16 +62,19 @@ public function removePaymentMethods(PaymentMethodRouteResponse $originalData, S
...
 
-        try {
-            $cart = $this->getCart($context);
-        } catch (MissingCartServiceException $e) {
-            $this->logger->error($e->getMessage(), [
-                'exception' => $e,
-            ]);
-            return $originalData;
-        }
+        if ($this->isCartRoute()) {
```
boxblinkracer, line 65: i had this isCartRoute in the subscription remover for instances
somehow it didnt work on storefront
so i moved to a isOrderRoute and a simple ELSE for everything else
that worked

maybe the difference is that 'store-api.payment.method' was missing ? i dont know
did you test it with pwa after this change?

daniel-memo-ict: the store-api.payment.method route has also been added to valid cart routes, so it should work now

PR #424, opened 2022-09-26
MOL-733: Invalid order ID: ''

smemoict:


smemoict:
Are Mollie Subscriptions also connected to a Mollie Order?

src/Subscriber/MollieOrderBuildSubscriber.php
```diff
@@ -0,0 +1,35 @@
...
+use Kiener\MolliePayments\Event\MollieOrderBuildEvent;
+
+class MollieOrderBuildSubscriber implements \Symfony\Component\EventDispatcher\EventSubscriberInterface
+{
+    public const METADATA_SHORT_TRANSACTION_ID_KEY = 'tid';
+
+    /**
+     * @inheritDoc
+     */
+    public static function getSubscribedEvents()
```
boxblinkracer, line 14: i think we can just add this to the core function? this is more for third party devs :)
also it should then be even visible in the unit tests which doesnt seem to be the case right now

smemoict: Could you elaborate what you mean please?

PR #425, opened 2022-09-26
MOL-871: Webhook mails always have UTC time

daniel-memo-ict:
Shopware reads the user's timezone from a cookie, and sets that in the twig environment. During webhook calls this cookie is unavailable, and will default to UTC.
To fix this we store the timezone in the order, and when webhooks are called we set this timezone on the twig environment.

### DIFF
--- a/src/Subscriber/WebhookTimezoneSubscriber.php
+++ b/src/Subscriber/WebhookTimezoneSubscriber.php
@@ -41,13 +41,12 @@ class WebhookTimezoneSubscriber implements EventSubscriberInterface
     public function __construct(
         TransactionService $transactionService,
         LoggerInterface    $logger
-    )
-    {
+    ) {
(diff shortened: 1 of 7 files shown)


src/Subscriber/WebhookTimezoneSubscriber.php
```diff
@@ -0,0 +1,101 @@
...
+
+        $route = $request->get('_route');
+        $routeParams = $request->get('_route_params');
+
+        $this->logger->debug("Starting Webhook Timezone Fixer", [
+            'route' => $route,
+            'routeParams' => $routeParams,
+        ]);
+
+        if ($route !== 'frontend.mollie.webhook') {
```
boxblinkracer, line 61: does that work with headless? store-api webhook?

PR #443, opened 2022-10-21
Fix ttypo

JoshuaBehrens:
Just a small ttypo about a t that missed its spot


boxblinkracer:
thank you @JoshuaBehrens 
means a lot

JoshuaBehrens:
Thank you for merging :)

PR #445, opened 2022-10-26
fix: Change method of checking if constant exists

aragon999:
and prevent a fatal error if the `handlerIdentifier` of the payment method is not a valid class.

We currently experience this in combination with the stripe plugin:
fatal error


boxblinkracer:
Hi @aragon999 
oh wow.. interesting

thank you for this important fix, every improvement counts :)
I've started the pipeline and will merge it once its green

PR #451, opened 2022-11-11
MOL-949: remove payment url custom field 

daniel-memo-ict:


src/Facade/MolliePaymentDoPay.php
```diff
@@ -229,9 +227,8 @@ public function startMolliePayment(string $paymentMethod, AsyncPaymentTransactio
         $this->updaterLineItemCustomFields->updateOrderLineItems($mollieOrder, $salesChannelContext);
 
 
-        # this condition somehow looks weird to me (TODO)
-        $checkoutURL = $orderCustomFields->getMolliePaymentUrl() ?? $orderCustomFields->getTransactionReturnUrl() ?? $transactionStruct->getReturnUrl();
-
+        # TODO: this condition somehow looks weird to me
+        $checkoutURL = $mollieOrder->getCheckoutUrl() ?? $transactionStruct->getReturnUrl();
```
boxblinkracer, line 231: i guess we can remove the TODO? :) 
still wondering, why is checkout url used or return URL? why 2 different ones? could you add it as comment if you know?

boxblinkracer (changes requested):
please see comments

src/Facade/MolliePaymentDoPay.php
daniel-memo-ict: I'm not sure but I'm gonna guess it's a fallback for a case that technically could happen, but practically never happens? `getCheckoutUrl()` technically can return null, even though in practice that never happens, and I think whoever wrote that added the Shopware return url as a fallback, on the off chance that it does happen.

boxblinkracer: checkout URL can be null on apple pay payments with a token
its indeed null or empty in that case because a redirect doesnt make sense.
also with earlier non-3d secure cards (which are deprecated anyway)
but i think the code is good. could you add the info with apple pay and empty checkoutURLs...that should be fine then
another thing that pops up is..checkout URL is empty once the paymment is completed. what if we would re-visit this page? in the earlier version it redirected us to our history data...will this still work? is it still a thing?

daniel-memo-ict: Comment has been updated.

> another thing that pops up is..checkout URL is empty once the paymment is completed. what if we would re-visit this page? in the earlier version it redirected us to our history data...will this still work? is it still a thing?

If the payment has been completed they shouldn't be able to come back here.

PR #457, opened 2022-11-18
MOL-967: Show Refund Error Message

ThLind:


src/Controller/Api/Order/RefundController.php
```diff
@@ -383,8 +383,7 @@ private function refundAction(string $orderId, string $orderNumber, string $desc
                 [
                     'success' => false,
                     'message' => $e->getMessage()
-                ],
-                500
+                ]
```
boxblinkracer, line 386: sorry, but we just cant change the status code, because it makes sense
and also people use it :)
we have to tell the admin to also extract the message from a 500 response

boxblinkracer (changes requested):
we must not merge this in here.
the 500 must not change as response status

src/Controller/Api/Order/RefundController.php
ThLind: I don't think it makes sense. A 500 Error indicates that theres a problem on the server, the request itself is correct and the same request can be tried again when the error on the server is resolved, but in this method a 500 is even responded when an orderId is missing and so the request isn't correct.
In the case of the ticket the precondition (transaction has to be paid to be refunded) isn't fulfilled, so it's neither a request nor a server error. 
The next is that the count of 500 responses is often monitored by hosters to check for server problems so this behavior could trigger false alerts.
I think a 200 with success false would be the logical response in this case, so perhaps the better compromise would be to only return a 500 when a real server error occured, only return a 200 with success false for this kind of errors and some 4xx error for the missing orderId.

PR #458, opened 2022-11-23
NTR: Ensure that `title` never contains whitespace only

fjbender:
It might happen that the salutation/title wrongfully contains whitespace only. This will lead to Mollie rejecting the order with 

```
Error executing API call (422: Unprocessable Entity): The ‘title’ field should not contain only whitespace characters.
```

To prevent this, we `trim()` the display name of the salutation.

### DIFF
--- a/tests/PHPUnit/Service/MollieApi/Builder/MollieOrderAddressBuilderTest.php
+++ b/tests/PHPUnit/Service/MollieApi/Builder/MollieOrderAddressBuilderTest.php
@@ -79,6 +79,19 @@ class MollieOrderAddressBuilderTest extends TestCase
     }
 
     /**
+     * This test verifies that the value of a salutation (title in Mollies API)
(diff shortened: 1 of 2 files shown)


tests/PHPUnit/Service/MollieApi/Builder/MollieOrderAddressBuilderTest.php
```diff
@@ -78,6 +78,19 @@ public function testBuildWithMissingSalutation(): void
         self::assertNull($addressData['title']);
     }
 
+    /**
+     * This test verifies that the value of a salutation (title in Mollies API)
+     * will never contain whitespace only.
+     */
+    public function testBuildWithWhitespaceSalutation(): void
```
boxblinkracer, line 85: nice! danke

boxblinkracer (approved):

PR #474, opened 2022-12-09
MOL-763: Disable form submit in ajax mode, since shopware already submits the form

BlackScorp:


src/Resources/app/storefront/src/mollie-payments/plugins/creditcard-components-sw64.plugin.js
```diff
@@ -287,7 +287,9 @@ export default class MollieCreditCardComponentsSw64 extends Plugin {
             }
         }
 
-        form.submit();
+        if (!window.csrf.enabled && window.csrf.mode !== 'ajax') {
```
boxblinkracer, line 290: and the token is correctly stored before the main request?

BlackScorp: not sure, but i do not get any csrf errors there, so i assume that everything is processed correcly

PR #481, opened 2022-12-27
MOL-1002: add rounding diff on full refund

BlackScorp:


src/Components/RefundManager/RefundData/RefundData.php
```diff
@@ -132,6 +132,18 @@ public function toArray()
             $hydratedOrderItems[] = $item->toArray();
         }
 
+        /** @var array<mixed> $refundsArray */
+        $refundsArray = $this->refunds;
+        foreach ($refundsArray as $refundIndex => $refund) {
+            if (isset($refund['metadata']['composition']) && is_array($refund['metadata']['composition'])) {
```
boxblinkracer, line 138: i see that we define a rounding item as swReference === false?
but wouldnt it make sense to use the metadata.type that exists for these items?
metadata.type === 'rounding'

src/Components/RefundManager/RefundManager.php
```diff
@@ -317,6 +320,30 @@ private function buildRequestItemsFromOrder(OrderEntity $order)
         return $items;
     }
 
+    private function appendRoundingItemFromMollieOrder(RefundRequest $request, ?\Mollie\Api\Resources\Order $mollieOrder): void
```
boxblinkracer, line 323: would it make sense to remove the nullable?
so that the caller needs to verify already, but also in this scope...do we really have nullable orders? wouldnt that be a problem already?

src/Struct/MollieApi/OrderLineMetaDataStruct.php
```diff
@@ -0,0 +1,65 @@
...
+        $this->type = $this->getTypeFromMetaData($orderLine->metadata, self::TYPE_LINE_ITEM);
+        $this->amount = (float)$orderLine->totalAmount->value;
+    }
+
+    private function getTypeFromMetaData(\stdClass $metaData, string $default): string
+    {
+        if (property_exists($metaData, 'type') === false) {
+            return $default;
+        }
+        return $metaData->type;
```
boxblinkracer, line 43: string cast?

boxblinkracer (changes requested):
see comments

src/Components/RefundManager/RefundManager.php
BlackScorp: Mollie Orders do not exists in Subscription

src/Components/RefundManager/RefundData/RefundData.php
BlackScorp: there is no type in the response from mollie API. We have metadata type which is "full" but the composition it self does not contain information about rounding diff, we have only swReference available

boxblinkracer (approved):

PR #494, opened 2023-01-27
fix: Reinitialization of the global Mollie instance

aragon999:
We have a one page checkout, where we replace the payment form, for example on login.

After the replacement we reinitialize all JavaScript plugins using
```javascript
window.PluginManager.initializePlugins();
```

However the Mollie credit card plugin throws an error:
grafik

As far as I understand one may only have one instance of the Mollie plugins, because of the loaded iframes and thus only one credit card form per page.

This pull request tries to improve the situation, by saving the Mollie instance in the `window` object and only allow one instance for each component type.

Note that this approach does not work when one has multiple credit card forms on the page.


boxblinkracer:
wow @aragon999 
thank you so much for this
well check it out some time in the next week.
if it doesnt break, we can already add it to the next version :)

thank you so much for this

boxblinkracer:
thank you @aragon999 
just tested it, looks good and will be merged :)

thank you for your help

PR #506, opened 2023-02-08
NTR: add debug log when no subscription was created

michnhokn:
@boxblinkracer Just wanted to add a debug message when no subscription was created.
Turns out we accidentally got rid of custom fields when persisting the cart, but mollie subscriptions depend on them. So it made sense for me to now have it in debug mode even if no subscription is created.

### DIFF
--- a/src/Components/Subscription/Actions/CreateAction.php
+++ b/src/Components/Subscription/Actions/CreateAction.php
@@ -46,6 +46,7 @@ class CreateAction extends BaseAction
 
         if (!$attributes->isSubscriptionProduct()) {
             # Mixed carts are not allowed for subscriptions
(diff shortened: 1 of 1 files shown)


boxblinkracer:
hey @michnhokn 
absoluteley, totally makes sense
thank you so much for this

i will do the cs-fixing for you and probably turn the message into "did not create subscription., etc....probably no subscription product." might be a good hint for other merchants with the "might not be a subscription product"

thank you

boxblinkracer:
oh and i will probably do a warning level, hope thats also ok

michnhokn:
@boxblinkracer Nope thats fine. Thank you for merging it so fast 😄

boxblinkracer:
perfect, thanks @michnhokn 
yeah..sometimes im pretty fast :)

PR #516, opened 2023-02-22
MOL-864 Full Refund with Flow Builder + Refund Manager after partial refund throws error

smemoict:


src/Components/RefundManager/RefundManager.php
```diff
@@ -300,25 +298,66 @@ private function getOrderItem(OrderEntity $orderEntity, string $lineID): ?OrderL
      */
     private function buildRequestItemsFromOrder(OrderEntity $order)
     {
+        $refunds = $this->refundService->getRefunds($order);
+
+        $refundCalculationHelper = new RefundCalculationHelper();
+        /** @var array<mixed> $refund */
+        foreach ($refunds as $refund) {
```
boxblinkracer, line 305: so i tried to figure out what this does,
i think its calculating what is already refunded to substract it?
so a) could you also add comments
and b) doesnt it look like the code form the RefundDataBuilder ? there is already such a logic to get the refunded values from composition and mollie line item quantities

could you double check and reuse that if so?

boxblinkracer (changes requested):
please see comments, thank you

src/Components/RefundManager/RefundManager.php
```diff
@@ -426,4 +465,29 @@ private function convertToRefundItems(RefundRequest $request, OrderEntity $order
 
         return $serviceItems;
     }
+
+    /**
+     * @param OrderDeliveryEntity|OrderLineItemEntity $lineItem
+     * @return string
+     */
+    public function getOrderLineId($lineItem): string
```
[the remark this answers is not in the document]
smemoict: Agreed, it has been integrated with existing code

src/Components/RefundManager/RefundManager.php
smemoict: Agreed, re-using.

boxblinkracer:
hey hey
thank you so much for this my friend 

i cant remember why it was never been finished, if it was our fault im so sorry
so we did it with the current version of the plugin, which means i would close this PR

if something is not working, just let us know :)

PR #517, opened 2023-02-22
Add composer 2 as requirement

daniel-memo-ict:


boxblinkracer:
fails on 6.3.5.0, that still allows composer 1, so we cannot change it for now, right?

boxblinkracer:
merged in other commit

PR #560, opened 2023-03-31
Fixes #553

### LINKED ISSUE #553: Critical Bug: Static quantity in buildShippingLineItems
Hi 

Currently, all orders are rejected if a shipping method is charged more than once. Here is an example: 
Bildschirmfoto 2023-03-30 um 10 38 22
--> 2x Zuschlag (1x = 2,50€)

We noticed that in the function "buildShippingLineItems" a static 1 is given. If e.g. 2x 2.50 € comes from Shopware, the Mollie API expects 2.50€, but gets a total of 5.00€ and runs into an error:
https://github.com/mollie/Shopware6/blob/master/src/Service/MollieApi/Builder/MollieShippingLineItemBuilder.php

```
  $mollieLineItem = new MollieLineItem(
          OrderLineType::TYPE_SHIPPING_FEE,
          sprintf('Delivery costs %s', $i),
          1,
          $price,
          $delivery->getId(),
          sprintf('mol-delivery-%s', $i),
          '',
          ''
      );
```

This should be fixed as soon as possible. It's really critical. We have a lot of cancellations and the customer cannot order with any Mollie payment method. 

We are losing a lot of big orders here right now.

I would appreciate a quick feedback!

JanMue-dot:
Fixes the Issue [#553](https://github.com/mollie/Shopware6/issues/553)

### DIFF
--- a/src/Service/MollieApi/Builder/MollieShippingLineItemBuilder.php
+++ b/src/Service/MollieApi/Builder/MollieShippingLineItemBuilder.php
@@ -43,6 +43,7 @@ class MollieShippingLineItemBuilder
         foreach ($deliveries as $delivery) {
             $i++;
             $shippingPrice = $delivery->getShippingCosts();
+            $qty = $shippingPrice->getQuantity();
             $totalPrice = $shippingPrice->getTotalPrice();
 
             if ($totalPrice === 0.0) {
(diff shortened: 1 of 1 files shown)


boxblinkracer:
hey @JanMue-dot , ok thats a nice one, didnt know aggregated qty is in that place :)
great job, thx

Issue #579, opened 2023-05-31
Custom field installation failing on "uniq.custom_field_set_relation.entity_name" constraint


Lapsus:
When installing the plugin on a fresh Shopware 6 instance the custom fields are not created properly.
During debugging i found a constraint violation on "uniq.custom_field_set_relation.entity_name" in the CustomFieldService class.

It looks as the relation custom field set -> customer entity was added twice.

I will also add a PR that fixes the problem by removing the duplicate relation.

PR #580, opened 2023-05-31
fixing adding duplicate relation during installation of custom fields

Lapsus:
fixed https://github.com/mollie/Shopware6/issues/579

### DIFF
--- a/src/Service/CustomFieldService.php
+++ b/src/Service/CustomFieldService.php
@@ -35,16 +35,17 @@ class CustomFieldService
     public function addCustomFields(Context $context): void
     {
         try {
+            $fieldSetId = Uuid::randomHex();
             $mollieOrderFieldId = Uuid::randomHex();
             $mollieCustomerFieldId = Uuid::randomHex();
             $iDealIssuerFieldId = Uuid::randomHex();
 
             $this->repoCustomFieldSets->upsert([[
-                'id' => Uuid::randomHex(),
+                'id' => $fieldSetId,
                 'name' => 'mollie_payments',
                 'config' => [
                     'label' => [
-                        'en-GB' => 'Mollie'
+                        'en-GB' => 'Mollie',
                     ]
                 ],
                 'customFields' => [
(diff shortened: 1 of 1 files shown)


Lapsus:
Shure, 

i just install the plugin via the admin ui or the command line in a plain, fresh Shopware 6.
And if i check the custom fields after the plugin installation and activation only the mollie_payments_product custom field set is added. The mollie_payments field set and the fields the CustomFieldsService should add, are missing. 

I then startet to dig into the problem and was able to finde the unique constraint violation caused by adding the two relations for the field set and the customer entity.

boxblinkracer:
Hi

alright thank you
got it..so i dont know that CustomFieldService....I only did the CustomFieldsInstaller
the class you have touched uses random hex ids...wouldnt that recreate new fields over and over on additional installations? 
thats why i did add fixed IDs for upserts in the other class

i have to re-check this :)
did you try to install a couple of times? do they exist multiple times then?

boxblinkracer:
i get this error with your code when uninstalling and installing again in the admin

Screenshot 2023-05-31 at 20 19 42

Lapsus:
I always use a blank installation and have not tried to install/uninstall multiple times. But this could be fixed i think by either removing the fieldset on uninstall, or by checking for existing fieldset before reinstalling.

boxblinkracer:
Ok i got it
there was a try catch, so this wasnt even inserted, and im not quite sure if its even good to have it :)
so ill check it out, thanks a lot for this

Lapsus:
frankly i am not sure how this ever worked in the first place. 
i found checked the problem was introduced with version 1.2 i think, when the third field (ideal prefered bank) was added.

boxblinkracer:
yeah but that should also work without a custom field set. just plain json storing. and we also have custom displays in the admin. so if that upsert did never work. then nobody needs it 🥹

Lapsus:
removed code is good code ;) 
maybe the best solution in that case to remove the class completely.
Would also solve the install/uninstall issue.

BlackScorp:
hey @Lapsus thank you for the PR and showing the issue. as discussed with @boxblinkracer  everything is clear now so i can close the issue.

PR #607, opened 2023-07-31
MOL-1196: Tracking

BlackScorp:


src/Facade/MollieShipment.php
```diff
@@ -478,6 +488,11 @@ private function createTrackingInfoStruct(string $trackingCarrier, string $track
             throw new \InvalidArgumentException('Missing Argument for Tracking Code!');
         }
 
+        $trackingUrl = trim(sprintf($trackingUrl, $trackingCode));
+        if (filter_var($trackingUrl, FILTER_VALIDATE_URL) === false) {
+            $trackingUrl = '';
```
boxblinkracer, line 493: please see sw5 plugin, theres more like % placeholders and stuff i can rememeber

src/Service/MollieApi/Order.php
```diff
@@ -391,14 +394,17 @@ public function getPaymentUrl(string $mollieOrderId, string $salesChannelId): ?s
...
-
+        $shipment = [];
+        if ($trackingInfoStruct instanceof ShipmentTrackingInfoStruct) {
+            $shipment['tracking'] = $trackingInfoStruct->toArray();
+        }
         /** @var OrderLine $orderLine */
         foreach ($mollieOrder->lines() as $orderLine) {
             if ($orderLine->shippableQuantity > 0) {
-                $mollieOrder->shipAll();
+                $mollieOrder->shipAll($shipment);
```
boxblinkracer, line 407: sure that not providing it an an empty array is handled in the same way?
just double checking

src/Service/TrackingInfoStructFactory.php
```diff
@@ -0,0 +1,54 @@
...
+{
+    public function createFromDelivery(OrderDeliveryEntity $orderDeliveryEntity):?ShipmentTrackingInfoStruct
+    {
+        $trackingCodes = $orderDeliveryEntity->getTrackingCodes();
+        $shippingMethod = $orderDeliveryEntity->getShippingMethod();
+        if ($shippingMethod === null) {
+            return null;
+        }
+        if (count($trackingCodes) !== 1) {
+            return null;
```
boxblinkracer, line 19: mhm why are mutliple codes not allowed?
could you add a comment for that please

src/Service/TrackingInfoStructFactory.php
```diff
@@ -0,0 +1,54 @@
+<?php
+declare(strict_types=1);
+
+namespace Kiener\MolliePayments\Service;
+
+use Kiener\MolliePayments\Struct\MollieApi\ShipmentTrackingInfoStruct;
+use Shopware\Core\Checkout\Order\Aggregate\OrderDelivery\OrderDeliveryEntity;
+
+class TrackingInfoStructFactory
```
boxblinkracer, line 9: this class is missing a few things that i can remember from shopware 5
https://github.com/mollie/Shopware/blob/master/Components/Mollie/MollieShipping.php

so tracking codes have a maximum length, we need to truncate these like in sw5
carrier can be missing which is ok, in this case we need to send a "-"

src/Service/TrackingInfoStructFactory.php
```diff
@@ -0,0 +1,54 @@
...
+
+        if (empty($trackingCarrier)) {
+            throw new \InvalidArgumentException('Missing Argument for Tracking Carrier!');
+        }
+
+        if (empty($trackingCode)) {
+            throw new \InvalidArgumentException('Missing Argument for Tracking Code!');
+        }
+
+        $trackingUrl = trim($trackingUrl . $trackingCode);
```
boxblinkracer, line 39: this does not work, thats just a string concat
in sw5 there was a placeholder being allowed for specific values
so we need to figure out if its the same, or come up with something like {trackingCode}, and then this would need to replaced, a basic concat leads to an invalid url very likely

tests/PHPUnit/Service/TrackingInfoStructFactoryTest.php
```diff
@@ -0,0 +1,46 @@
...
+        $this->factory = new TrackingInfoStructFactory();
+    }
+
+    /**
+     * @dataProvider invalidCodes
+     * @param string $url
+     * @param string $trackingCode
+     * @return void
+     */
+    public function testInvalidTrackingCodeCharacter(string $trackingCode):void{
```
boxblinkracer, line 30: quite sure a lot of cases are missing in here, also positive ones
here is a list of possible test cases
https://github.com/mollie/Shopware/blob/master/Tests/PHPUnit/Components/Mollie/Shipping/MollieShippingTest.php

src/Facade/MollieShipment.php
```diff
@@ -250,7 +261,7 @@ public function shipOrder(
         $shipment = $this->mollieApiShipmentService->shipOrder(
             $mollieOrderId,
             $order->getSalesChannelId(),
-            $this->createTrackingInfoStruct($trackingCarrier, $trackingCode, $trackingUrl)
+            $this->trackingInfoStructFactory->create($trackingCarrier, $trackingCode, $trackingUrl)
```
boxblinkracer, line 264: really wondering, this PR is about automating tracking in whatever shipping is being done.
so if we extract tracking from delivery in the constructor....then i would assume its used somewhere, but i only see function arguments being used here, and the calling instance doesnt seem to have changed which means its the old logic? so where is the auto-extracted data then used?

boxblinkracer (changes requested):
comments

src/Service/TrackingInfoStructFactory.php
BlackScorp: in SW6 there is trackingCode variable just in email, the URL itself does not have a tracking Code

https://docs.shopware.com/en/shopware-6-en/tutorials-and-faq/track-and-trace  it seems like sw6 expects that the tracking code is at the end of the URL

src/Facade/MollieShipment.php
BlackScorp: the shipment facade is too big and too much things, the tests are too complicated, i wanted to extract some private methods into seperate services in order to write smaller tests. the factory is not used anywhere else but it makes the tests easier

tests/PHPUnit/Service/TrackingInfoStructFactoryTest.php
BlackScorp: the positives ones are tested here

https://github.com/mollie/Shopware6/blob/8f63450b225e13541fed965e4b538104a33c6706/tests/PHPUnit/Facade/MollieShipment/SetShipmentTest.php

src/Service/TrackingInfoStructFactory.php
boxblinkracer: nope, the info icon says clearly its the placeholder %s

tests/PHPUnit/Service/TrackingInfoStructFactoryTest.php
boxblinkracer: i only see entity written tests or missing-xyz-tests but no positive ones

src/Facade/MollieShipment.php
boxblinkracer: this doesnt really give an answer to the question
i think we should talk about it :)

PR #611, opened 2023-08-22
Allow salesChannelId to be missing

JoshuaBehrens:
nwdxxItRechtConnector seems to create an entry without the salesChannelId, which is okay to do, but I get

```
[ErrorException]
Warning: Undefined array key "salesChannelId"
```


boxblinkracer:
Hey @JoshuaBehrens 
wohooo, perfect, totally fine with that, thanks for your great help :)

boxblinkracer:
thanks :)

PR #619, opened 2023-09-14
Expose refund amount to flow.

kleinmann:
This allows the flow builder to access the refund amount, e.g. for customer mails.

Example in a mail template:
```
{{ amount | currency(order.currency.isoCode) }}
```


boxblinkracer:
nice one @kleinmann 
thanks a lot :)

Issue #507, opened 2023-02-09
Reduce amount of API Call when switching payment methods


fjbender:
What is currently happening?

Whenever I switch payment methods in the frontend, the Mollie Profile and Methods APIs are called.

![image](https://user-images.githubusercontent.com/7370694/217791337-ac4faa75-351a-4010-9f3e-47e275e263df.png)

What should be happening?

The content of the API calls is fairly static and can safely be cached, at least within one checkout session. The API should not be called that often.

boxblinkracer:
huch? that is not known :D 
nice finding, we should fix that

fjbender:
This issue got worse after the POS payment method has been introduced. Now `/v2/methods/ideal` and `/v2/terminals` are called for every payment method selection screen, even when those payment methods are turned off.

BlackScorp:
It is live now and we should have see less requests

PR #659, opened 2023-11-24
MOL-1263: load data for selected payment methods only

BlackScorp:
Here were some Problems

1) ideal issues, pos terminal and credit card mandate were loaded on every request
2) profile id was loaded twice, one general and one for ideal issues

now:
1) we load only additional data if mollie payments were selected
2) the profile id is loaded once and saved for the request
3) ideal, pos, CC data is loaded if the payment methods are selected


fjbender:
Fixes #507

boxblinkracer (approved):

PR #678, opened 2024-01-10
MOL-1287: do not overwrite exisisting technical name

BlackScorp:


src/Service/PaymentMethodService.php
```diff
@@ -183,19 +184,15 @@ public function addPaymentMethods(array $paymentMethods, Context $context): void
...
 
-                # starting with Shopware 6.5.7.0 this has to be filled out
-                # so that you can still save the payment method in the administration
-                if ($this->versionCompare->gte('6.5.7.0')) {
-                    $paymentMethodData['technicalName'] = self::TECHNICAL_NAME_PREFIX . $mollieName;
-                }
-
-                $upsertData[] = $paymentMethodData;
+                # we do a string cast here, since getTechnicalName will be not nullable in the future
+                $technicalName = (string)$existingPaymentMethod->getTechnicalName();
```
boxblinkracer, line 188: i think you can just use $paymentMethodData['technicalName'] and it should be the same

src/Service/PaymentMethodService.php
```diff
@@ -209,15 +206,19 @@ public function addPaymentMethods(array $paymentMethods, Context $context): void
...
                 ];
+            }
 
-                # starting with Shopware 6.5.7.0 this has to be filled out
-                # so that you can still save the payment method in the administration
-                if ($this->versionCompare->gte('6.5.7.0')) {
-                    $paymentMethodData['technicalName'] = self::TECHNICAL_NAME_PREFIX . $mollieName;
-                }
+            if (mb_strlen($technicalName) > 0) {
+                $technicalName = self::TECHNICAL_NAME_PREFIX . $mollieName;
```
boxblinkracer, line 215: doenst it make sense to place it in the if condition so that its all together with that feature, now its decentralized

src/Service/PaymentMethodService.php
```diff
@@ -183,19 +184,15 @@ public function addPaymentMethods(array $paymentMethods, Context $context): void
                     'name' => $existingPaymentMethod->getName(),
                 ];
 
-                # starting with Shopware 6.5.7.0 this has to be filled out
-                # so that you can still save the payment method in the administration
-                if ($this->versionCompare->gte('6.5.7.0')) {
```
boxblinkracer, line 188: i think technical name for inserts does not work anymore?

src/Service/PaymentMethodService.php
boxblinkracer: breaks in shopware < 6.5.6

src/Service/PaymentMethodService.php
BlackScorp: i moved the  line    $upsertData[] = $paymentMethodData;  outside of if else for that

boxblinkracer (changes requested):
see comments

src/Service/PaymentMethodService.php
BlackScorp: i changed that

boxblinkracer (approved):

PR #683, opened 2024-01-15
MOL-1280: Fix retry route and add cypress tests

BlackScorp:
i know that the code is repeated, we have to comeup here with a better architecture, i would suggest to move if to refactoring after release?

### DIFF
--- a/tests/Cypress/cypress/e2e/storefront/subscriptions/subscription.cy.js
+++ b/tests/Cypress/cypress/e2e/storefront/subscriptions/subscription.cy.js
@@ -76,61 +76,8 @@ describe('Subscription', () => {
 
             describe('Storefront + Administration', function () {
                 it('C2339889: Purchase subscription after failed payment and verify data in Administration', () =>{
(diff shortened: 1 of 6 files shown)


tests/Cypress/cypress/e2e/storefront/subscriptions/subscription.cy.js
```diff
@@ -75,6 +75,120 @@ describe('Subscription', () => {
...
+                    cy.contains('.btn', 'Subscribe');
+                    // we also want to see the translated interval
+                    cy.contains('Every 3 weeks');
+
+                    pdp.addToCart(2);
+
+                    // ------------------------------------------------------------------------------------------------------
+
+                    // verify our warning information in our offcanvas
+                    cy.contains('Not all payments methods are available when ordering subscription products');
```
boxblinkracer, line 99: i think you can remove a few checks in here, maybe keep line 90 with the contains of button Subscribe.
this would help to only focus on the current test in here, and have those "backbone-logic" tests in the actual core cypress test (purchase subscription....)

just an idea

PR #700, opened 2024-02-08
Update register.html.twig

uthiess:
Insert hidden field to let Shopware know we want to create an account. Without that field set, checkout will always continue as guest.


fjbender:
Thank you for your contribution! Is that true in all cases, has it changed recently? I remember testing this a while back and it wasn't an issue. 😅

uthiess:
I'm not sure since it only caught my attention a few days ago. I believe it worked fine when we tested it around Q3 of last year.

PR #721, opened 2024-03-12
NTR: updates for 6.6

BlackScorp:


akf-bw:
@BlackScorp Any news about the release of the 6.6 update? It's been almost a month since Shopware 6.6 was officially released and this is one of the biggest payment provider plugins in the Shopware Store. The update is needed.

BlackScorp:
@resslinger @akf-bw iam working on it, as you can see in the pull request, we had to change all routes from annotations to xml so the plugin still works in shopware 6.4 and 6.5, i have to test everything before i can make the release. the pipeline is still red.

akf-bw:
@BlackScorp I think releasing multiple versions is no longer possible. 6.6 plugins are not compatible due to JavaScript structure. (https://github.com/shopware/shopware/blob/trunk/UPGRADE-6.6.md#re-compile-your-javascript) Other plugin developers have created a second branch for the legacy versions. (if something needs to be fixed for older versions)

BlackScorp:
@akf-bw we had no problems with javascript, since we dont have that much javascript in the frontend anyways, compiling worked without problems for all version. Also the Javascript part works so far.

you can download the plugin zip from here https://github.com/mollie/Shopware6/actions/runs/8468889420 to check it out

BlackScorp:
@akf-bw we use custom webpack to build it, the js file is then included only if its needed via twig files.

https://github.com/mollie/Shopware6/blob/master/src/Resources/views/storefront/base.html.twig#L39

the filesize for our plugin is 8.1 KB 

we have still some old stuffs from Shopware 6.2 times (including proper usage of shopware js) a clean up will follow but not a priority atm

akf-bw:
@BlackScorp by looking into the JS structure I found something that should be improved: https://github.com/mollie/Shopware6/pull/736
Currently the browser has to wait until the full parse of the mollie javascript before he can continue parsing the page html.

resslinger:
@BlackScorp everything looks green now, great!

BlackScorp:
@resslinger now the QA part beginns, have to check all the edge cases, since the entire routing was rewritten

PR #790, opened 2024-07-09
PISHPS-328: added ApplePayDirectDomainAllowList

m-muxfeld-diw:
added ApplePayDirectDomainAllowList, integrated allow list into applepay direct validate route, updated documentation


src/Resources/config/config.xml
```diff
@@ -152,6 +152,15 @@
                 </option>
             </options>
         </input-field>
+        <input-field type="text">
+            <name>ApplePayValidationAllowList</name>
```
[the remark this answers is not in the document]
boxblinkracer: and please use applePayDirect as prefix, because non-direct would be another payment method

src/Resources/config/config.xml
```diff
@@ -152,6 +152,15 @@
...
+            <label>Allowed Domains for Apple Pay Validation</label>
+            <label lang="de-DE">Zulässige Domains für die Apple Pay Validierung</label>
```
boxblinkracer, line 158: could you add something like "leave empty if no separate headless ...stuff with additional domain is used"
or "only required if a custom domain is used for a headless storefront"
just to avoid people getting confused and thinking this needs to be added
one of the mollie usps is a plugn'play approach without lots of required configs

src/Components/ApplePayDirect/ApplePayDirect.php
```diff
@@ -473,6 +491,29 @@ public function createPayment(OrderEntity $order, string $shopwareReturnUrl, str
         return $paymentData->getMollieID();
     }
 
+    /**
+     * @param string $validationUrl
+     * @throws ApplePayValidationUrlAllowListCanNotBeEmptyException
+     * @throws ApplePayValidationUrlNotInAllowListException
+     * @return string
+     */
+    public function validateValidationUrl(string $validationUrl): string
```
boxblinkracer, line 500: validate DOMAIN? or?
validation URL is something else, we need to validate the domain
which also means we should rename the whole member variables and probably classes?
(we can talk in zoom if you want)

src/Resources/config/config.xml
m-muxfeld-diw: currently it is always required but the ApplePayConfigSubscriber adds all associated domains of the shop automatically to the configuration on install/update

src/Components/ApplePayDirect/ApplePayDirect.php
boxblinkracer: also why do we require a string return value? its no "getter" it just validates with exception approach, meaning void is totally fine?
so sanitizing might be better as a second function? function naming and content is a bit different i think?

src/Components/ApplePayDirect/Gateways/ApplePayValidationUrlAllowListGateway.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * Get the ApplePayValidationUrlAllowList
+     *
+     * @return ApplePayValidationUrlAllowList
+     */
+    public function getAllowList(): ApplePayValidationUrlAllowList
+    {
+        $allowList = $this->systemConfigService->get('MolliePayments.config.ApplePayValidationAllowList');
```
boxblinkracer, line 34: please use our own SettingsService to avoid duplicate hardcoded config strings

src/Components/ApplePayDirect/Subscriber/ApplePayConfigSubscriber.php
```diff
@@ -0,0 +1,110 @@
...
+     */
+    private $salesChannelDomainRepository;
+
+    /**
+     * ApplePayConfigSubscriber constructor.
+     *
+     * @param SystemConfigService $systemConfigService
+     * @param EntityRepository $salesChannelDomainRepository
+     */
+    public function __construct(SystemConfigService $systemConfigService, EntityRepository $salesChannelDomainRepository)
```
boxblinkracer, line 40: i dont see any benefit in this?
we get the string anyway if we use the  settings service when we need this?!

tests/PHPUnit/Components/ApplePayDirect/Gateways/ApplePayValidationUrlAllowListGatewayTest.php
```diff
@@ -0,0 +1,36 @@
...
+    {
+        $this->service->expects($this->once())->method('get')->willReturn('');
+        $allowList = $this->gateway->getAllowList();
+
+        $this->assertTrue($allowList->isEmpty());
+    }
+
+    public function testProvidesAllowList(): void
+    {
+        $allowListString = 'https://example.com,https://example-url.org';
```
boxblinkracer, line 29: great test
could you also add a simple domain or a separate unit test where it loads just domains without https scheme?

tests/PHPUnit/Components/ApplePayDirect/Models/Collections/ApplePayValidationUrlAllowListTest.php
```diff
@@ -0,0 +1,41 @@
...
+namespace MolliePayments\Tests\Components\ApplePayDirect\Models\Collections;
+
+
+use Kiener\MolliePayments\Components\ApplePayDirect\Models\ApplePayValidationUrlAllowListItem;
+use Kiener\MolliePayments\Components\ApplePayDirect\Models\Collections\ApplePayValidationUrlAllowList;
+use PHPUnit\Framework\TestCase;
+
+class ApplePayValidationUrlAllowListTest extends TestCase
+{
+    public function canDetermineIfValueIsContainedInList(): void
```
boxblinkracer, line 13: attention
tests without "test" prefix will not be executed

tests/PHPUnit/Components/ApplePayDirect/Services/ApplePayValidationUrlSanitizerTest.php
```diff
@@ -0,0 +1,37 @@
...
+    {
+        $sanitizedUrl = $this->sanitizer->sanitizeValidationUrl($url);
+
+        $this->assertEquals($expected, $sanitizedUrl);
+    }
+
+    public function sanitationTestDataProvider(): array
+    {
+        return [
+            'keeps http value if provided' => ['http://example.com', 'http://example.com/'],
```
boxblinkracer, line 30: not quite sure if trailing slash is good?

tests/PHPUnit/Service/ApplePayDirect/ApplePayDirectTest.php
```diff
@@ -131,6 +145,49 @@ public function testBuildApplePayCart(): void
...
     }
 
+    public function testThrowsExceptionWhenAllowListIsEmpty(): void
+    {
+        $this->validationUrlAllowListGateway->expects($this->once())
+            ->method('getAllowList')
+            ->willReturn(ApplePayValidationUrlAllowList::create());
+
+        $this->expectException(ApplePayValidationUrlAllowListCanNotBeEmptyException::class);
+        $this->expectExceptionMessage('The Apple Pay validation URL allow list can not be empty. Please check the configuration.');
```
boxblinkracer, line 155: thats wrong, it can be empty

boxblinkracer (changes requested):
comments

src/Components/ApplePayDirect/Gateways/ApplePayDirectDomainAllowListGateway.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * Get the ApplePayDirectDomainAllowList
+     *
+     * @param SalesChannelContext $context
+     * @return ApplePayDirectDomainAllowList
+     */
+    public function getAllowList(SalesChannelContext $context): ApplePayDirectDomainAllowList
+    {
+        $settings = $this->settingsService->getSettings($context->getSalesChannel()->getId());
+        $allowList = $settings->applePayDirectDomainAllowList ?? '';
```
boxblinkracer, line 37: achtung, hier sollte ein getter das schon machen, direkt member-access sollte nicht erlaubt sein

src/Components/ApplePayDirect/Models/ApplePayDirectDomainAllowListItem.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public static function create(string $value): self
+    {
+        if (empty($value)) {
+            throw new \InvalidArgumentException(sprintf('The value of %s must not be empty', self::class));
+        }
+
+        # we need to have a protocol before the parse url command
+        # in order to have it work correctly
+        if (strpos($value, 'http') !== 0) {
```
boxblinkracer, line 37: i think you could use the sanitizer in here?

src/Service/SettingsService.php
```diff
@@ -18,6 +18,7 @@ class SettingsService implements PluginSettingsServiceInterface
     const TEST_API_KEY = 'testApiKey';
     const LIVE_PROFILE_ID = 'liveProfileId';
     const TEST_PROFILE_ID = 'testProfileId';
+    const APPLE_PAY_DIRECT_DOMAIN_ALLOW_LIST = 'applePayDirectDomainAllowList';
```
boxblinkracer, line 21: glaub das ist nicht notwendig

boxblinkracer: aaaa, MollieSettingsStruct is the correct one

boxblinkracer (changes requested):
comments

tests/PHPUnit/Components/ApplePayDirect/Gateways/ApplePayDirectDomainAllowListGatewayTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $this->service = $this->createMock(SettingsService::class);
+        $this->gateway = new ApplePayDirectDomainAllowListGateway($this->service);
+        $this->scContext = $this->createConfiguredMock(SalesChannelContext::class, [
+            'getContext' => $this->context = $this->createMock(Context::class),
+        ]);
+    }
+
+    public function testProvidesEmptyAllowList(): void
+    {
+        $struct = $this->createConfiguredMock(MollieSettingStruct::class, ['getApplePayDirectDomainAllowList' => '']);
```
boxblinkracer, line 49: just a styling thing, but maybe something for the future
you dont always need to mock things
you could just create a new setting struct, set your value and use it :)

tests/PHPUnit/Components/ApplePayDirect/Services/ApplePayDirectDomainSanitizerTest.php
```diff
@@ -0,0 +1,39 @@
...
+    {
+        $sanitizedUrl = $this->sanitizer->sanitizeDomain($url);
+
+        $this->assertEquals($expected, $sanitizedUrl);
+    }
+
+    public function sanitationTestDataProvider(): array
+    {
+        return [
+            'removes http value if provided' => ['http://example.com', 'example.com'],
```
boxblinkracer, line 30: can you add a subdomain too? that would be good

boxblinkracer (approved):

PR #799, opened 2024-08-05
fix: load order delivery state machine state

niklaswolf:
The change adds loading of the state-machine state to the loading of the order in `OrderDeliveryService`.
This is necessary, because otherwise the technical name of the state machine state can't be loaded here: https://github.com/mollie/Shopware6/blob/9558e0de87d0bc43190ae69242e9b1d77eb3e47b/src/Service/Transition/DeliveryTransitionService.php#L92 

This leeds to two delivery-state transitions: from shipped to open (see line 106) and then back to shipped (l. 109).
The second and third marked transition in the screenshot are caused by this bug.
Bildschirmfoto 2024-08-05 um 14 05 52

I patched the file like in this PR, now it's working as intended


fjbender:
Thank you for your contribution, @niklaswolf!

PR #806, opened 2024-08-09
OrderEditSubscriber

m-muxfeld-diw:
Reopens unconfirmed orders to enable the end user to make another attempt at paying

### DIFF
--- a/makefile
+++ b/makefile
@@ -81,7 +81,11 @@ phpmin: ## Starts the PHP compatibility checks
 	@php vendor/bin/phpcs -p --standard=PHPCompatibility --extensions=php --runtime-set testVersion 7.4 ./src
 
 csfix: ## Starts the PHP CS Fixer
+ifndef fix
 	@PHP_CS_FIXER_IGNORE_ENV=1 php vendor/bin/php-cs-fixer fix --config=./.php_cs.php --dry-run
+else
+	@PHP_CS_FIXER_IGNORE_ENV=1 php vendor/bin/php-cs-fixer fix --config=./.php_cs.php
+endif
 
 stan: ## Starts the PHPStan Analyser
 	@php vendor/bin/phpstan analyse -c ./.phpstan.neon
--- a/src/Service/Order/OrderTimeService.php
+++ b/src/Service/Order/OrderTimeService.php
@@ -1,0 +1,56 @@
+<?php
+declare(strict_types=1);
+
+namespace Kiener\MolliePayments\Service\Order;
+
+use DateTime;
+use Shopware\Core\Checkout\Order\Aggregate\OrderTransaction\OrderTransactionEntity;
+use Shopware\Core\Checkout\Order\OrderEntity;
+
+class OrderTimeService
+{
+    /**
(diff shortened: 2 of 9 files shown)


makefile
```diff
@@ -81,7 +81,11 @@ phpmin: ## Starts the PHP compatibility checks
 	@php vendor/bin/phpcs -p --standard=PHPCompatibility --extensions=php --runtime-set testVersion 7.4 ./src
 
 csfix: ## Starts the PHP CS Fixer
+ifndef fix
```
BlackScorp, line 84: can you remove this part? we have the command make pr which fixes everything

src/Service/Order/OrderTimeService.php
```diff
@@ -0,0 +1,56 @@
...
+class OrderTimeService
+{
+    /**
+     * @var DateTime
+     */
+    private $now;
+
+    public function __construct(?DateTime $now = null)
+    {
+        $this->now = $now ?? new DateTime('now');
```
BlackScorp, line 19: 'now' is default value anyways, please remove

src/Subscriber/OrderEditSubscriber.php
```diff
@@ -0,0 +1,132 @@
...
+                continue;
+            }
+
+            $transactions = $order->getTransactions();
+
+            if ($transactions === null || $transactions->count() === 0) {
+                continue;
+            }
+
+            $lastTransaction = $transactions->last();
```
BlackScorp, line 65: here we might have the problem that last transaction is not last by date. the transactions are sorted by id, we might use here a filter method or something similar so we will have the latest transaction by date

src/Subscriber/OrderEditSubscriber.php
```diff
@@ -0,0 +1,132 @@
...
+
+            $lastStatus = $lastTransaction->getStateMachineState()->getTechnicalName();
+
+            // disregard any orders that are not in progress
+            if ($lastStatus !== 'in_progress') {
```
BlackScorp, line 70: please avoid magic values, we have a class with status constants please use them instead

src/Subscriber/OrderEditSubscriber.php
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        return $paymentMethod->getHandlerIdentifier() === BankTransferPayment::class;
+    }
+
+    private function isMolliePayment(OrderEntity $order): bool
+    {
+        $customFields = $order->getCustomFields();
+
+        return is_array($customFields) && !empty($customFields) && isset($customFields['mollie_payments']);
```
BlackScorp, line 130: please avoid empty since empty is true if $customFields is false or '0'. empty can produce more bugs, better use count

BlackScorp (changes requested):

src/Service/Order/OrderTimeService.php
m-muxfeld-diw: now is only the default value in production. In the test class other DateTime's are injected into the class

PR #821, opened 2024-09-03
PISHPS-304: added OrderTagService

m-muxfeld-diw:


src/Service/Tags/OrderTagService.php
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        // Fetch or create the tag
+        $criteria = new Criteria();
+        $criteria->addFilter(new EqualsFilter('id', $subscriptionTag->getId()));
+        /** @var null|TagEntity $tag */
+        $tag = $this->tagRepository->search($criteria, $context)->first();
+
+        if (!$tag instanceof TagEntity) {
+            throw CouldNotTagOrderException::forSubscription(sprintf('Tag with name "%s" and ID "%s" not found', $subscriptionTag->getName(), $subscriptionTag->getId()));
```
BlackScorp, line 64: can you just log the info instead of throwing exception here? sometimes customers likes to delete tags, just in case that buying subscription still will work without tags

src/Service/Tags/OrderTagService.php
```diff
@@ -0,0 +1,93 @@
...
+        /** @var null|OrderEntity $order */
+        $order = $this->orderRepository->search($criteria, $context)->get($orderId);
+
+        if (!$order instanceof OrderEntity) {
+            throw CouldNotTagOrderException::forSubscription(sprintf('Order with ID "%s" not found', $orderId));
+        }
+
+        // Fetch or create the tag
+        $criteria = new Criteria();
+        $criteria->addFilter(new EqualsFilter('id', $subscriptionTag->getId()));
```
BlackScorp, line 59: can you use here the tag name, merchants can delete and create tags on the fly

src/Service/Tags/OrderTagService.php
```diff
@@ -0,0 +1,93 @@
...
+            ],
+        ], $context);
+    }
+
+    /**
+     * @return array<string, string>
+     */
+    private function serializeTag(TagEntity $tag): array
+    {
+        return ['id' => $tag->getId(),];
```
BlackScorp, line 91: is this comma intentional?

BlackScorp (changes requested):

src/Service/Tags/OrderTagService.php
m-muxfeld-diw: It was, but I don't mind removing it

PR #867, opened 2024-10-17
PISHPS-368: Hide cart apple pay option

ThLind:


src/Resources/views/storefront/page/checkout/cart/index.html.twig
```diff
@@ -21,6 +35,7 @@
 
 {% block page_checkout_aside_actions %}
     {{ parent() }}
+    {{ dump() }}
```
MelvinAchterhuis, line 38: Oops 😄

PR #868, opened 2024-10-21
use more specific way to check if a backup cart is existing

netzkollektiv:
We're having problems with the loading behavior of the `Kiener\MolliePayments\Service\Cart\CartBackupService`. Our plugin is dynamically adding an additional `LineItem` to the current cart using a custom `Collector`. However, the `LineItem` is also being added to the backup cart which leads to the problem that `isBackupExisting` returns `true` without an existing backup cart. 

It would be good if the Mollie extension could use a more specific method to check if a backup cart exists. I would suggest the use of `Shopware\Core\Checkout\Cart\CartPersister::load(string $token, SalesChannelContext $context), see this pull request.

### DIFF
--- a/src/Resources/config/services.xml
+++ b/src/Resources/config/services.xml
@@ -130,6 +130,7 @@
 
         <service id="Kiener\MolliePayments\Service\Cart\CartBackupService" class="Kiener\MolliePayments\Service\Cart\CartBackupService" public="true">
             <argument type="service" id="Shopware\Core\Checkout\Cart\SalesChannel\CartService"/>
+            <argument type="service" id="Shopware\Core\Checkout\Cart\CartPersister"/>
         </service>
 
         <service id="Kiener\MolliePayments\Service\Mollie\OrderStatusConverter" public="false">
--- a/src/Service/Cart/CartBackupService.php
+++ b/src/Service/Cart/CartBackupService.php
@@ -19,13 +19,18 @@ class CartBackupService
      */
     private $cartService;
 
+    /**
+     * @var AbstractCartPerister 
+     */
+    private $cartPersister;
 
     /**
      * @param CartService $cartService
      */
-    public function __construct(CartService $cartService)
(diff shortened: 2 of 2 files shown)


BlackScorp:
hi @netzkollektiv 
thx for the pull request, maybe we can check why the line items also beeing added to the backup cart? maybe there is a problem? 

right now the backup cart is created when you click on apple pay direct button. there i send a value to add to cart port request. if the value is not set, the backup cart should not be created in first place.

if the backup cart is also filled with your collector, then the there will be no exception there so the method isBackupExisting will still return true, right?

netzkollektiv:
Hi @BlackScorp, thank you for your quick feedback ! Indeed we are going to adapt our extension and check if the cart token is the current token within the collector. However I‘m not really sure how the collectors are meant to be. For me they seem to be implemented in a „singleton way“. Regardless of our change, I think the explicit check in the isBackupExisting check would avoid problem like that in general.

thanks, Dominik

netzkollektiv:
@BlackScorp 
Your change does not solve the problem, the `LineItem` would be added anyway. 

The cart persister executes just one simple query which will give you a definite answer to your question if the cart is existing. If you mind about performance you could simplify and execute the query yourself: `SELECT 1 FROM cart WHERE `token` = :token`  

see the original db statement here:
https://github.com/shopware/shopware/blob/c20e50cdeec0c06ad40f2680562aa955a4e5ce87/src/Core/Checkout/Cart/CartPersister.php#L37

BlackScorp:
@netzkollektiv i think i understand, getCart creates a new cart if one is not existing. i will add a select statement with in memory cache

BlackScorp:
@netzkollektiv i could not use the connection i have to use the persister, since there is also a redis cart persister and others can configure which one is used.

i had to change some things as well, so i created a new PR

https://github.com/mollie/Shopware6/pull/875

i will close this PR then. hope then your problem is fixed

netzkollektiv:
Thank you very much for your efforts! :-)

PR #883, opened 2024-10-30
NTR - optimizes custom domain overwrite

michnhokn:
I would suggest adapting the `applyCustomDomain` method to replace not just the host but the full base URL. This way you can also control the schema as well as any ports. 

In the current implementation you might run into the problem that you have a local port like `8000` which is not the case for the URL you want to replace it with. Then only the host is replaced, and Mollie will still try to access the new host on port `8000`, which is not what you want.

This changes the behaviour of the `MOLLIE_SHOP_DOMAIN` env variable a bit, because it should provide a schema. But it's not necessary.


BlackScorp:
hi @michnhokn thanks for the PR, could also fix the Unit tests for this? then i would merge it.

best regards

PR #947, opened 2025-02-20
NTR: avoid misleading logs in payment remover route detection. fix #563

### LINKED ISSUE #563: Error created by Bingbot?
I have these errors in the Mollie log:

Looks like its created by Bingbot?

afbeelding

content of the error:
`Could not determine if the current route is allowed to remove payment methods (Session: ...) {"exception":"[object] (Kiener\\MolliePayments\\Exception\\MissingRouteException(code: 0): Could not determine the route at /home/h100525/domains/DOMAIN.nl/public_html/custom/plugins/MolliePayments/src/Service/Payment/Remover/PaymentMethodRemover.php:215)","request":{"Symfony\\Component\\HttpFoundation\\Request":"GET /atom.xml HTTP/1.1\r\nAccept: */*\r\nAccept-Encoding: gzip, deflate\r\nCache-Control: no-cache\r\nForwarded: for=\"52.167.XXX.XX\";host=\"DOMAIN.nl\";proto=https\r\nFrom: bingbot(at)microsoft.com\r\nHost: DOMAIN.nl\r\nPragma: no-cache\r\nSw-Context-Token: lQDzK7wiCLPYzndUQRriNHPquAZppWNB\r\nSw-Language-Id: 389d33f672524d0aa38d1dd521f8eb1a\r\nUser-Agent: Mozilla/5.0 AppleWebKit/537.36 (KHTML, like Gecko; compatible; bingbot/2.0; +http://www.bing.com/bingbot.htm) Chrome/103.0.5060.134 Safari/537.36\r\nX-Forwarded-For: 52.167.XXX.XX\r\nX-Php-Ob-Level: 1\r\n\r\n"},"session":"","processors":{"uid":{"uid":"520XXX523"},"web":{"url":"/atom.xml","ip":"52.167.XXX.X","http_method":"GET","server":"DOMAIN.nl","referrer":null},"introspection":{"file":null,"line":null,"class":"Kiener\\MolliePayments\\Service\\Payment\\Remover\\PaymentMethodRemover","callType":"->","function":"isAllowedRoute"}}} []`

boxblinkracer:
so we had the problem of /atom.xml leading into exceptions and logs (bing-bot)

somehow it seems as if custom configurations lead to resolving of routes that shouldnt be resolved

then i think a bot detection or anything like this could lead to the route attribute being missing

nevertheless....our main goal here is just check if the route matching our expected cart or order routes
in that case it actually does not

so i think we should not have an impact in the exception chain in this case

so ive changed to fail-safe approaches

also removed the optional parameters because that leads to additional complexity and intertwining

### DIFF
--- a/src/Service/Payment/Remover/CartAwareRouteInterface.php
+++ b/src/Service/Payment/Remover/CartAwareRouteInterface.php
@@ -12,7 +12,7 @@ interface CartAwareRouteInterface
         'store-api.payment.method',
     ];
 
-    public function isCartRoute(string $route = ""): bool;
+    public function isCartRoute(): bool;
 
     public function getCart(\Shopware\Core\System\SalesChannel\SalesChannelContext $context): \Shopware\Core\Checkout\Cart\Cart;
 }
--- a/src/Service/Payment/Remover/OrderAwareRouteInterface.php
+++ b/src/Service/Payment/Remover/OrderAwareRouteInterface.php
@@ -8,7 +8,7 @@ interface OrderAwareRouteInterface
         'frontend.account.edit-order.page',
     ];
 
-    public function isOrderRoute(string $route = ""): bool;
+    public function isOrderRoute(): bool;
 
     public function getOrder(\Shopware\Core\Framework\Context $context): \Shopware\Core\Checkout\Order\OrderEntity;
 }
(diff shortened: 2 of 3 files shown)


BlackScorp:
@boxblinkracer nice, but then we dont need the exception at all right? we can remove the cartch statement and the exception since it is not thrown anyways

boxblinkracer:
yes that exception might be gone. but theres often a chance for other exceptions and the fail safe approach should still exit to avoid troubles in the shop

PR #1046, opened 2025-07-23
PISHPS-505: fix for webhooks which are triggered too early

### LINKED ISSUE #1037: Bug: order_transaction.state.open -> paid is called twice
I recently updated to Shopware 6.7 and mollie/Shopware 4.16 and have the problem, that the webhook and the checkout both change the state from "open" to "paid".
This leads to problems, since I send a custom email to the shop-owner on this event, which is now sent twice.

I read something about "updatePayment race conditions" in the release-info, maybe that is related.

Attached a screenshot of the shopware-logs

Image

uehler:
fix #1037

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 
+### Fixed
+- Fixed webhook race condition
+
 ## [4.16.0] - 2025-06-18
 
 ### Added
--- a/src/Components/ApplePayDirect/ApplePayDirect.php
+++ b/src/Components/ApplePayDirect/ApplePayDirect.php
@@ -428,6 +428,7 @@ class ApplePayDirect
             $paymentData->getMollieID(),
             '',
             $transaction->getId(),
+            0,
             $context->getContext()
         );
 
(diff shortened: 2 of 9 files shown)


RobinHoutevelts:
@BlackScorp I might be overlooking something, but can you explain how this fixes #1037?

It looks like we're now keeping track of how often we've received the webhook. But I can't seem to find where we \*use\* that counter to prevent the event from being dispatched multiple times.

A `sleep(2)` has been added (which I expect hurt high-traffic instances btw), is this to prevent the race condition? If so, shouldn't there be a check immediately after the sleep to see if we've already processed this particular webhook?

Thanks for looking into this btw 🙏

BlackScorp:
@RobinHoutevelts the code which is preventing the status change, was there 
https://github.com/mollie/Shopware6/blob/master/src/Service/Order/OrderStatusUpdater.php#L88

we never change a status if the status has reached already the target. the idea is, call the method as much as you want, if the status is already there, there should be no status change. as you can see from the code, right before status change, we verify ther we need a status check.

the issue here is that the redirect back and the webhook is executed at same time. so either the status was not changed yet in the database or the result is cached somehow.

we are speaking right now with mollie about an argument where we can define when a webhook should be send. i think the best solution would be, if i send a timestamp to mollie api when a webhook should be send. 

since mollie does not provide such feature we have try out different things. the counter is only there so we have a sleep timer for the very first transaction. afterwards we dont need it anymore.


we are aware of the sleep counter that its not the perfect solution. the solution we had before with too early exception was not perfect either. other merchants were confused about the errors in the webhook logs in mollie dashboard, which increased unnecessary support tickets. 

So the options we have right now
1) add a sleep timer
2) remove status change after redirect
3) remove status change in webhook

if we would add 2) then development and testing the plugin is really hard, since my local environment must accept webhooks and also all test server need this as well.

if we would remove it in 3) then some payment methods would never change their status. e.g. some banking payment methods works with a QR Code. you scan the QR Code on your smartphone and then there is no redirect back to shop. so status is changed only via webhook.

thats the problems we are currently facing. i think that the sleep timer is the best solution right now with the possitibilities i have. once mollie provides a timestamp parameter when the first webhook should be send, then we would have a perfect solution.

unless you have better ideas, maybe iam missing something.

PR #1161, opened 2025-12-12
Update price check to include zero value

damian-pastorini:
When the cart is empty the price is zero and because of that the $availableMolliePaymetnts returns empty. This cause issue where you won't be able load the payments using the SW PaymentMethodRoute. By included the zero condition you will get the original data.


BlackScorp:
@damian-pastorini no need, thanks for the pull request

PR #1239, opened 2026-03-30
PISHPS-574: Change saving card

BlackScorp:


fjbender (changes requested):

fjbender:
This will not yet work in test mode, but the `payment` payload in the create order call will look something like this:

```
  "payment": {
    "customerId": "cst_NaVGCqHyTf",
    "storeCredentials": true,
    "cardToken": "tkn_Rnz7Btest",
    "webhookUrl": "https://example.com/webhook"
  },
```

fjbender (approved):

PR #1254, opened 2026-04-29
NTR: add one off to sequence type

BlackScorp:


src/Handler/Method/CreditCardPayment.php
```diff
@@ -49,6 +49,7 @@ public function processPaymentMethodSpecificParameters(array $orderData, OrderEn
             // change payment sequenceType to first if this is a single-click payment
```
fjbender: This comment seems outdated

PR #1280, opened 2026-06-17
Fix typo in Slovene Translation (sl_SL -> sl_SI)

gecolay:
There is a typo in the Slovene Translation file:
- Currently: "sl_SL" which is no valid ISO code
- Correct: "sl_SI"

I reaplaced all occurrences of "sl_SL" & "sl-SL" with "sl_SI" & "sl-SI"

The latest Shopware 6.7.11.0 no longer works with the latest Mollie Version, as now ISO codes of all snippets are correctly validated, so this has to be fixed for Mollie to work with the latest Shopware version asap.


BlackScorp:
hi @gecolay thanks for the pull request,

i will check it out with the new version. right now it is not supported in the plugin store, so we havent tested our plugin against the new version yet.

BlackScorp:
fix #1281

BlackScorp:
@gecolay could you also update the path in this file as well?

https://github.com/mollie/Shopware6/blob/master/config/.phpunuhi.xml

thanks in advance

gecolay:
@BlackScorp This file is already updated: https://github.com/mollie/Shopware6/pull/1280/changes#diff-0efd1ae96f3b8f3adfc6f186f23f1af873078646a136af96380324b804499f8b
Or am I missing something?

PR #1348, opened 2026-07-07
fix: map line items with negative price to discount type (fixes #1347)

### LINKED ISSUE #1347: v5.0.0 regression: third-party discount line items are sent as 'digital' — Mollie rejects payment (unitPrice cannot be negative)
### Plugin version
5.0.0

### Shopware version
6.6.10.1

### PHP version
8.3

### Description
Since v5.0.0 payments fail with a 422 from the Mollie API whenever the cart contains a discount line item that was added by a third-party plugin (i.e. any line item type other than Shopware's core `promotion`/`credit` types), for example voucher or customer-discount plugins:

```
Mollie Pay Process Failed
{"error":"Error in field lines.10.unitPrice. Unprocessable Entity: Line item 10 is invalid.
The 'unitPrice' field cannot be negative, unless the 'type' is one of the following:
'discount', 'store_credit', and 'gift_card'","paymentMethod":"paypal"}
```

### Root cause
`Mollie\Shopware\Component\Mollie\LineItemType::fromOderLineItem()` only maps the core Shopware line item types:

```php
return match ($oderLineItemType) {
    LineItem::PRODUCT_LINE_ITEM_TYPE, self::LINE_ITEM_TYPE_CUSTOM_PRODUCTS->value => self::PHYSICAL,
    LineItem::CREDIT_LINE_ITEM_TYPE => self::CREDIT,
    LineItem::PROMOTION_LINE_ITEM_TYPE => self::DISCOUNT,
    default => self::DIGITAL,
};
```

Every custom line item type falls through to `digital`. If such a line has a negative price (a discount), the Mollie API rejects the whole payment.

### Regression from v4
v4 handled this case explicitly in `MollieLineItemBuilder`:

```php
if ($item->getType() === $this->getLineItemPromotionType() || $item->getTotalPrice() < 0) {
    return OrderLineType::TYPE_DISCOUNT;
}
```

Any line item with a negative total price was sent as `discount`. This logic was lost in the v5 rewrite, so shops using any third-party discount/voucher plugin cannot take payments through Mollie for affected carts.

### Steps to reproduce
1. Install any plugin that adds a discount as a line item with a custom type (cart processor adding a `LineItem` with a negative `CalculatedPrice`).
2. Place an order with that discount and pay with any Mollie payment method.
3. The payment fails with the 422 above; the customer sees a payment error.

### Expected behaviour
Line items with a negative total price should be sent to Mollie as type `discount` (unless they are already `store_credit`/`gift_card`), like v4 did.

I have a fix ready and will open a PR right away.

marcmanusch:
Fixes #1347

## Problem
`LineItemType::fromOderLineItem()` maps every line item type it does not know to `digital`. Discounts added by third-party plugins (vouchers, customer discounts, …) use custom line item types, so they are sent to the Mollie API as `digital` with a negative `unitPrice` — the API rejects the whole payment:

```
Error in field lines.10.unitPrice. Unprocessable Entity: Line item 10 is invalid.
The 'unitPrice' field cannot be negative, unless the 'type' is one of the following:
'discount', 'store_credit', and 'gift_card'
```

This is a regression from v4, where `MollieLineItemBuilder` explicitly mapped any line item with a negative total price to `TYPE_DISCOUNT`:

```php
if ($item->getType() === $this->getLineItemPromotionType() || $item->getTotalPrice() < 0) {
    return OrderLineType::TYPE_DISCOUNT;
}
```

## Fix
`fromOderLineItem()` now returns `DISCOUNT` for any line item with a negative total price, unless the type already resolved to `store_credit` or `gift_card` (which Mollie also accepts with negative amounts). The price is read via `getPrice()` (nullable, safe on partially hydrated entities) instead of the non-nullable `totalPrice` property.

Shipping lines are unaffected — negative shipping costs are already handled in `LineItem::fromShippingCosts()`, and the `RoundingDifferenceFixer` already sets `DISCOUNT` for negative rounding lines.

## Tests
Added two tests to `tests/Unit/Mollie/LineItemTypeTest.php`:
- negative custom-type, negative product and negative credit line items map to `discount` / `store_credit`
- positive custom-type line items still map to `digital`

```
OK (9 tests, 18 assertions)   # LineItemTypeTest
OK, but there were issues!    # full tests/Unit: 818 tests, 2116 assertions (pre-existing PHPUnit notices only)
```

php-cs-fixer (config/.php_cs.php) and phpstan pass on the changed files.

Verified against a real failed order (net tax status, cart promotion + third-party voucher line): the voucher line now resolves to `discount`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG_de-DE.md
+++ b/CHANGELOG_de-DE.md
@@ -10,6 +10,7 @@
 - Behoben: Die Storefront bricht auf Shopware 6.5 nicht mehr mit dem Fehler „Plugin is already registered" ab.
 - Behoben: Die Anzeige-Einschränkungen für Apple Pay Direct greifen nun, sodass der Button auf den konfigurierten Seiten ausgeblendet wird.
 - Behoben: Storefront-Seiten brechen bei von Mollie nicht unterstützten Locales (z. B. cs_CZ, sk_SK) nicht mehr ab. Die Locale weicht nun auf eine unterstützte oder auf en_GB aus.
+- Behoben: Zahlungen schlagen nicht mehr fehl, wenn der Warenkorb einen Rabatt eines Drittanbieter-Plugins enthält (eigener Positionstyp mit negativem Preis). Solche Positionen werden nun als Typ 'discount' an Mollie übertragen.
 
 # 5.0.0
 - Hinweis: Durch Autoloader-Caching kann beim Hochladen/Update des Plugins ein Fehler erscheinen. Dieser kann ignoriert werden.
--- a/CHANGELOG_en-GB.md
+++ b/CHANGELOG_en-GB.md
@@ -10,6 +10,7 @@
 - Fixed: The storefront no longer breaks with a "Plugin is already registered" error on Shopware 6.5.
 - Fixed: Apple Pay Direct display restrictions are now applied, so the button is hidden on the configured pages.
 - Fixed: Storefront pages no longer break on locales that Mollie does not support (e.g. cs_CZ, sk_SK). The locale now falls back to a supported one, or to en_GB.
+- Fixed: Payments no longer fail when the cart contains a discount from a third-party plugin (custom line item type with a negative price). Such line items are now sent to Mollie as type 'discount'.
 
 # 5.0.0
 - Note: Due to autoloader caching, an error can appear when uploading/updating the plugin. It can be ignored.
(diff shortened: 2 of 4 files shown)


BlackScorp:
@marcmanusch thanks for this! could you also update the changelogs?

marcmanusch:
Done — added entries to CHANGELOG_en-GB.md and CHANGELOG_de-DE.md under "Unreleased".

PR #1350, opened 2026-07-07
NTR: new feature payment link

### LINKED ISSUE #275: Mollie Checkout URL as a variable for Email Templates
It would be real nice to have the mollie checkout url for a customer as a variable for email templates. The regular Shopware routine for failed payments is not the best one in terms of conversion and we've received better results with directly sending customers the mollie checkout url.

### LINKED ISSUE #542: Feature Request: Send payment link from admin
Hi,

We often make manual orders and would be great if we can send an payment link directly from the admin when making the order. 

Would be great if this can be bases on an SW template, so we can modify the layout.

The 'reminder' function us used for failed payments from online orders.

BlackScorp:
fix #542 
fix #275


BlackScorp:
will try a different approach

PR #1365, opened 2026-07-13
NTR: fix payment with additonal path

BlackScorp:


konstantinkoeppe:
Hi Vitalij, it's Konstantin from the Shopware ticket - here are the details, since the formatting works better on this side.

I went through your changes against our installation (MolliePayments 5.1.0, Shopware 6.7.12.1). The two return-flow fixes look right to me:

- **`RouteBuilder` using `router` instead of `router.default`** is exactly the fix we applied on our side on Monday - we override that same constructor argument via a compiler pass, since we cannot patch the plugin itself.
- **Forwarding with the inherited routing attributes plus `sw-skip-transformer`** is the correct fix. It is what `StorefrontController::forwardToRoute()` does, so the sub-request keeps the sales channel and finalize actually runs. It also makes our own workaround (a decorator that let a forwarded sub-request inherit the main request's routing attributes instead of throwing) unnecessary, which is the outcome we were hoping for.

## One issue in the new paid-guard

In `Pay::execute()`:

```php
return new RedirectResponse($shopwareFinalizeUrl);
```

`$shopwareFinalizeUrl` is core's `payment.finalize.transaction` URL for the transaction that is *currently being paid*. `PaymentProcessor::getCurrentOrderTransaction()` only ever hands `pay()` a transaction in the **initial** state, never the paid one. So whenever the guard triggers, it redirects into finalize for a transaction that has no Mollie payment attached, because you return before `savePaymentExtension()`.

Finalize then runs into this:

1. `getPaymentByTransactionId()` finds no `mollie_payments` extension on the transaction
2. it falls back to `repairLegacyTransaction()`, which needs the order custom fields `order_id` and `transactionReturnUrl`. Those are v4-era keys: `Payment::toArray()` serialises the struct's camelCase property names (`orderId`, `finalizeUrl`), and nothing in 5.x writes `transactionReturnUrl` at all, so it returns `null`
3. -> `TransactionWithoutMollieDataException`
4. -> `AbstractMolliePaymentHandler::finalize()` wraps it as `PaymentException::asyncFinalizeInterrupted`
5. -> core's `PaymentProcessor::finalize()` catches it and calls `transactionStateHandler->fail()`

So the double charge is prevented, which is the main thing. But the already-paid order ends up with a **failed transaction as its newest one**, the payment-failed flows fire on an order that is in fact paid, and the customer is sent to the token's error URL (in the account flow that is the edit-order page, i.e. a payment form again) instead of the finish page.

Redirecting to `frontend.checkout.finish.page` for the order - the way `PaymentController::redirectAfterFinalize()` already handles the consumed-token case - would avoid this. Cancelling the redundant open transaction rather than leaving it behind would be the icing on the cake.

## Two things the guard does not cover

In case they are not already on your list:

1. The guard only checks paid/authorized **at creation time**. If a customer abandons the payment at Mollie, retries, and then completes the first payment from the still-open tab, both payments are chargeable: the earlier one is never cancelled, and in our case the card payments came back as `isCancelable: false`. That is how one of our orders ended up with three completed payments.
2. `savePaymentExtension()` overwrites the stored Mollie payment id, and `WebhookRoute::notify()` resolves the payment through `getPaymentByTransactionId()`, i.e. from that single field rather than from the payment id Mollie sends in the webhook body. Once a retry has happened, the earlier payment can never reconcile.

The `PrefixDomainFixture` and the Cypress test are a very welcome addition. That path-prefixed domain is exactly the condition that hides this class of bug from a default installation.

BlackScorp:
@konstantinkoeppe 

thanks for your details, ill will check the flow.

According to your questions.

1) i tried to handle the issue with this Pull request https://github.com/mollie/Shopware6/pull/1353/changes.

I checked shopware how shopware shows the payment status in the order overview. i thought first, that shopware just orders the transactions by date and show the latest. but it looked like that shopware shows in the overview the latest successfull transactions. all failed transactions afterwards are added to the database but not shown in the admin.

so if a customer completes both payments and will add two multiple payments, then he will be charged twice that is true. but the issue here is, that i cannot cancel open payments via API, i already tried it. payments with status open is cancelled. 

so i cannot prevent the customer for paying multiple times.

i also tried to store the inital payment id and reuse it, on a 2nd payment attempt. but here the api does not allow to change the payment method. you have to stick to the same payment method until the payment is either paid or cancelled. 

since this is a very rare case, that a customer opens multiple tabs and then complete the payment in multiple tabs, we decided not to handle that. 

right now i have only the option to reuse the payment id from mollie and if a customer started the checkout with paypal and then opens the shop in a new tab and edit the order, that i hide all other payment methods so he cannot switch it untill he actively cancelled it on the side of paypal. but this would confuse the customer. 

2) as far as i understood. paymentExtensions are primarly saved per transaction in the custom fields. i added additionally the data to order custom fields, they are not used by our plugin, the fields are used for external plugin like JTL this plugin can only read from order custom fields to export transaction data to a Warenwirtschaftsystem :D 

So every transactions maps to one payment at mollie, and every payment has his own webhook URL for transaction. 

I will check your description and see what i can do. but cancel a payment or prevent the customer from paying mulitple times is not possible for me.

BlackScorp:
@konstantinkoeppe i opened a new issue to discuss further details https://github.com/mollie/Shopware6/issues/1375. it is easier than searching for closed tickets

PR #1410, opened 2026-07-23
Only check mollie order's for changed delivery state

gecolay:
- The subscriber fails at `repairLegacyTransaction`, if there are order deliveries from other payment providers which do not have mollie data at all
- We now check, if it is a mollie order before using the ShipOrderRoute

### DIFF
--- a/shopware/Component/Shipment/OrderDeliverySubscriber.php
+++ b/shopware/Component/Shipment/OrderDeliverySubscriber.php
@@ -4,13 +4,17 @@ declare(strict_types=1);
 
 namespace Mollie\Shopware\Component\Shipment;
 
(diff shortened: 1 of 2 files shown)


BlackScorp:
hi @gecolay thanks for the pull request, could you update the changelog as well and check why the unit test fails?

you can just run following in your shop

cd custom/plugins/MolliePayments
composer install
make phpunit

this will install the dependencies and run phpunit, maybe you could fix that as well so the pipeline will be gree.

Thanks in advance.

gecolay:
@BlackScorp I see you already fixed the pipeline and merged the change, thanks for the fast response.

PR #1414, opened 2026-07-24
Fix OrderFileHandler incorrect log level & increase non debug log level to warning

gecolay:
- The OrderFileHandler currently doesn't use the correct log level from the plugin settings, but always debug right now
- With a disabled debug mode in the settings the log level should be `warning` instead of `info` as the `info` log level spam's the logs a lot

### DIFF
--- a/CHANGELOG_de-DE.md
+++ b/CHANGELOG_de-DE.md
@@ -1,5 +1,7 @@
 # Unreleased
 - Geändert: In Headless-Setups werden die Store-API-Webhook- und Return-URLs nun aus APP_URL statt aus dem Proxy-Host gebildet, damit Mollie sie erreichen kann.
+- Geändert: Das Log-Level wurde für den deaktivierten Debug-Modus der Einstellungen wurden von `INFO` zu `WARNING` geändert, um Spam in Logdateien zu verhindern.
+- Behoben: Das Log-Level für Order-bezogene Logeinträge verwendet nun das korrekte Log-Level der Debug-Modus Einstellung (`DEBUG` oder `WARNING`).
 - Behoben: Die Erfassung beim Versand sendet für Netto-Kunden nun den Bruttobetrag.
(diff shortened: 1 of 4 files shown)


BlackScorp:
hi @gecolay thanks for the pull request.

i will rebuild the pr a little bit. changing to warning in the code is not a good idea, i need the logs from info as well for debugging purpose. if a merchant did not enabled debug mode but wants to know what is happend with his failed order i can see there at least which process was triggered and when the error was there. it is helpfull for debugging for me.

i will rewrite the handler, so you can configure a loglevel like a regular log handler from monolog. and debug mode will overwrite your settings.

i will make a new pr okay?

BlackScorp:
@gecolay now you can adjust this over your shops config.yaml /config.xml see

https://github.com/mollie/Shopware6/pull/1417

gecolay:
> changing to warning in the code is not a good idea, i need the logs from info as well for debugging purpose.

> it is helpfull for debugging for me.

Isn't that what the "debug mode" in the settings is for? Mollie currently extremly spams the logs in prod, especially since the new structure, where a separate log file is created for each order.

> if a merchant did not enabled debug mode but wants to know what is happend with his failed order

A error will still show up in the logs, otherwise the merchant would have to enable the debug mode.


> @gecolay now you can adjust this over your shops config.yaml /config.xml see

I guess this will work for me (I still think that without debug mode, it should at least be a warning level, otherwise the logs grow enormously.)

BlackScorp:
@gecolay the issue is, that not every shop have enabled debug mode. and usually an issue happens and THEN they activate it and ask me what went wrong.

so i add info logs and warning etc and debug mode is for more detailed logs. it helps me with the support. not all merchants have a developer sitting with them some merchants are normal users and they ask me over support chat what went wrong, ineed a way to tell them even if the debug logs are disabled.

the spam is intended. we had issues where workers were not enabled in shop, was hard to debug without logs. we had issues where a shop was not reachable from outside so webhook didnt come in. its hard to debug if you dont have logs. so now i have many logs and i can tell then when something does not work.

PR #1416, opened 2026-07-24
Skip legacy order transaction payment method updates in webhook flow

gecolay:
- Right now the `notify` function of the `WebhookRoute` which e.g. runs in the scheduled task every minute fails, because there are could be still legacy orders in the system, which do not have the `method` field set in their custom fields
- The payment method update should then be skipped for this legacy order transactions

**Edit**: Cancelled transactions also have to mollie payment method set in the custom fields, we should also skip this transactions instead of throwing an exception. I updated the PR with this change

### DIFF
--- a/CHANGELOG_de-DE.md
+++ b/CHANGELOG_de-DE.md
@@ -3,7 +3,7 @@
 - Behoben: Die Erfassung beim Versand sendet für Netto-Kunden nun den Bruttobetrag.
 - Behoben: Anreden mit mehr als 20 Zeichen brechen die Zahlungserstellung nicht mehr ab; der Adress-Titel wird nun auf Mollies 20-Zeichen-Limit gekürzt.
 - Behoben: Der automatische Versand beim Wechsel des Lieferstatus läuft nun nur noch für über Mollie bezahlte Bestellungen.
-- Behoben: Das Aktualisieren der Zahlungsart im Webhook-Flow wird nun korrekt für alte Bestellungen übersprungen, statt eines Fehlers wegen fehlender Daten zu werfen.
+- Behoben: Das Aktualisieren der Zahlungsart im Webhook-Flow wird nun korrekt für alte Bestellungen und abgebrochene Bestellungen übersprungen.
 
 # 5.3.0
 - Hinzugefügt: Bestellungen können über einen Mollie-Payment-Link (Route `mollie.pay`) bezahlt werden, nutzbar in E-Mail-Templates, z. B. `{{ rawUrl('mollie.pay', { 'orderId': order.id }, salesChannel.domains|first.url) }}`.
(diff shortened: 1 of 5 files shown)


BlackScorp:
hi @gecolay 

i think we need to verify this issue. The thing is in the Mollie Gateway iam loading legacy orders and fix them.

https://github.com/mollie/Shopware6/blob/master/shopware/Component/Mollie/Gateway/MollieGateway.php#L97

PAyment without method exception is when a mollie payment does not have a specific custom field. maybe there is there issue and not legacy orders?

gecolay:
@BlackScorp Legacy order's can go through https://github.com/mollie/Shopware6/blob/master/shopware/Component/Mollie/Gateway/MollieGateway.php#L80 as they have the custom fields & the id

BlackScorp:
@g

> @BlackScorp Legacy order's can go through https://github.com/mollie/Shopware6/blob/master/shopware/Component/Mollie/Gateway/MollieGateway.php#L80 as they have the custom fields & the id

but thats okay. at the end there is a payment object which have a payment and for legacy there is also an order id.

maybe you can share your order_transaction custom fields so i can understand the issue better. for me it looks like that an old payment which is still fetchable over the API does not have a method name set. or the method name is set but it is none of our enum.

i createad a pr which is something like your but with some log information

https://github.com/mollie/Shopware6/pull/1418

gecolay:
> maybe you can share your order_transaction custom fields so i can understand the issue better. for me it looks like that an old payment which is still fetchable over the API does not have a method name set.

That’s exactly right. As far as I know, the "method" field didn't exist prior to the 5.0 update (at least, I never saw it in any order transaction).
So, regardless of the path taken within the MollieGateway, the Mollie API itself returns a "Payment" object without the "method" and it's also not in the custom fields.
So a call of the updatePaymentMethod method in the WebhookRoute will always fail right now for this legacy orders.

There are also new order transaction (from after the 5.0 update) in the system from orders where the payment flow was cancelled by the customer, there is also no "method" field in the mollie api or in the custom fields, so even new order transactions from the latest plugin version can fail right no every 60 seconds the scheduled task runs.

gecolay:
> i createad a pr which is something like your but with some log information

Your new PR includes all the features from my PR, so it should work.

Is there a release date for the new update yet?

BlackScorp:
@gecolay so method should return a method name, 

https://docs.mollie.com/reference/get-payment i assumed here that if set a method name during creation, then i would return one back when i fetch it..

anyways i didnt had this test case now i fixed this in the PR, thanks for pointing that out.

the release date will be in 3 weeks. iam on vacation for two weeks and when i come back i have to check what kind of new issues occured and fix/test them as well, so 3-4 weeks until next release. we have last 4 weeks every week a release.

if you need your changes right now, you can download the zip file from our actions https://github.com/mollie/Shopware6/actions/workflows/build_zip.yml

PR #1436, opened 2026-08-11
NTR: localize hardcoded storefront texts

cxo-jutz:
Several user- and screen-reader-facing strings were hardcoded in English. This adds proper snippet keys so they follow the shop language, matching the plugin's existing 23-locale setup.

**New snippet keys** (added to all 23 storefront locale files):
- molliePayments.closeModal — shared aria-label for modal close buttons (subscription modals ×2 and the mandate-remove modal)
- molliePayments.subscriptions.account.actionsMenuLabel — aria-label of the subscription actions menu button
- molliePayments.components.paypalExpress.checkout — visible label of the PayPal express button
- molliePayments.components.pos.returnToWebsite — POS return button (title + visible text)
- molliePayments.components.pos.terminalInstructions / paymentSuccess / paymentFailed / paymentFailedNoDebit — POS checkout status headlines shown via JS on the loading/success/failed states

**Template changes:**
- subscription-item.html.twig: actions-menu aria-label + both modal-close aria-labels now use |trans.
- credit-card-mandate-remove-modal.html.twig: modal-close aria-label now uses molliePayments.closeModal.
- paypal-express-button.html.twig: PayPal button label localized.
- pos/checkout.html.twig: return button and all three payment-status headlines localized. The page is rendered via renderStorefront() (route frontend.mollie.pos.checkout), so the storefront snippet set is loaded and |trans resolves.

**Not touched (intentional):** the test-mode-only Change Status in Test Mode link.
**Translations**: de-DE and en-GB are authoritative; the other 21 locales are best-effort and should get a native-speaker review before merge. Keys are structurally identical across all locales (camelCase, non-empty).


src/Resources/views/storefront/page/account/subscriptions/subscription-item.html.twig
```diff
@@ -537,7 +537,7 @@
                                                                                                     class="close"
                                                                                                     data-dismiss="modal"
                                                                                                     data-bs-dismiss="modal"
-                                                                                                    aria-label="Close">
```
BlackScorp, line 540: oh das ist ein guter fund, könntest du nicht schauen ob es nicht bereits snippets von shopware gibt wegen close?

BlackScorp (changes requested):

